PR #8, opened 2019-12-04
CI Update and Singularity Deployment

RobertRosca:
Updated `.travis.yml` and added in a singularity definition file just called `Singularity`.

The travis CI now has two stages: `test` and `singularity-build-push`.

The test section currently runs both the parallel and series tests for two versions of Python (3.7.5 and 3.6.9), with the same version of gcc (version 6, this could run for 6 and 7 but it would double the test time).

If the CI is running on the master branch or if a tag has just been made then the `singularity-build-push` stage will execute after the tests, this will build a singularity image and push it to the singularity registry.

The new image will be tagged either with the branch name or with the tag name. This way the `master` image will always give the latest version, and tagging a release acts to archive previous images.

Some examples from travis:

 - dev - only tests ran: https://travis-ci.com/RobertRosca/EXtra-foam/builds/139371279
 - master - triggers tests and image build/push: https://travis-ci.com/RobertRosca/EXtra-foam/builds/139372390
 - tag - triggers tests and image build/push: https://travis-ci.com/RobertRosca/EXtra-foam/builds/139372674

Images can be found [here](https://cloud.sylabs.io/library/_container/5de7ab59c9435e18ec5622e5) where you can see one tagged as `master`, and one tagged as `v0.0.0`.

Still have to:

 - Send Jun keys to add to Travis for the Singularity deployment stage
 - See if it's possible to use the European-XFEL organisation as an account on cloud.sylabs.io - as far as I can tell this can't be done, so the images will likely have to stay under my account for now
 - Add in help, labels, and runscript (maybe) to the image definition
 - Optional(?) add additional tests to the matrix, e.g. different versions of GCC, which act as a more rigorous test during merges to master
 - Optional(?) only enable some tests for merge to master

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -51,7 +51,7 @@ install:
   - python --version
   - which pip
   - pip --version
-  # QT_DEBUG_PLUGINS
+  # QT_DEBUG_PLUGINS 
   - if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then
       sudo apt install -y libxkbcommon-x11-0;
       pip install -e .[test];


.travis.yml
```diff
@@ -1,66 +1,34 @@
...
-      env: COMPILER=gcc GCC=7
+stages:
+  - test
+  - singularity-build-push
 
 env:
   global:
     - DISPLAY=":99.0"
-    - MINCONDA_VERSION="latest"
-    - MINCONDA_LINUX="Linux-x86_64"
```
zhujun98: I have those OS environment because I will have unittest for macos. I am not sure about windows since it will require a lot of effort on the C++ side. Will it also be easy to do this in your test workflow?

.travis.yml
```diff
@@ -1,66 +1,34 @@
...
+  jobs:
+    - PYTHON_VERSION="3.7.5"
+    - PYTHON_VERSION="3.6.9"
 
+services:
+  - xvfb
 before_install:
-  - |
-    if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then
-      if [[ "$COMPILER" == "gcc" ]]; then
```
zhujun98: I have this `if` because I also want to test compilers like `Clang` here. Nevertheless, the operational environment has `g++6` and `centos`. My first thought is that we may not want to put all the tests into Singularity because it is only meant for our operational environment. 

So the question is, can we separate them in `Travis`? i.e. a widely covered non-singularity test and a singularity test. Another immediate solution is to move former to the `Azure` pipeline.

.travis.yml
```diff
@@ -72,4 +40,58 @@ script:
...
+        - echo -e "$SYLABS_TOKEN" > token
+        - singularity remote login --tokenfile ./token || exit 1
+          #  key used for image signing
+          #  for travis ci pgp key block you must replace newline with '\\n', and replace spaces with '\ '
+        - echo -e "$SINGULARITY_KEY" > skey
+        - head -n 4 ./skey
+        - echo $SINGULARITY_KEY_PW | singularity key import ./skey || exit 1
+      script:
+        - cd $TRAVIS_BUILD_DIR
+        - export SINGULARITYENV_PYTHON_VERSION="3.7.5"
```
zhujun98: Why do we have a fixed Python version here?

Singularity
```diff
@@ -0,0 +1,88 @@
...
+  git clone https://github.com/European-XFEL/EXtra-foam
+  cd EXtra-foam
+  git checkout $GIT_COMMIT_HASH
+  git submodule update --init | tee -a $logpath/05-extra-foam-submodules.log
+  git log -n 1 | tee -a $logpath/06-extra-foam-clone.log
+
+  export FOAM_WITH_TBB=0
+  export XTENSOR_WITH_TBB=0
+  export FOAM_WITH_XSIMD=0
+  export XTENSOR_WITH_XSIMD=0
```
zhujun98: Why do you need those environmental variables here? By default they are all 1, which means to build the parallel version.

.travis.yml
```diff
@@ -72,4 +40,58 @@ script:
...
+        - export SINGULARITYENV_GIT_COMMIT_HASH=$TRAVIS_COMMIT
+        - sudo singularity build ./extra-foam.sif Singularity
```
zhujun98: Is there a better way or what is the encouraged way to name the singularity file? Here the command `singularity` only differs the filename `Singularity` by a capitalized `S`.

RobertRosca: I'll change the name to `singularity.def`, it was just `Singularity` before as I planned to build it using the official Singularity Hub builders, and those look for files formatted `Singularity.tag`, and as I'm not using the hub builder the file can be renamed to anything.

.travis.yml
RobertRosca: I can remove that and use `PYTHON_VERSION` instead, that way it'll take the first option from `env.jobs.PYTHON_VERSION`, which would be `PYTHON_VERSION="3.7.5"`.

So the version of python in the image will always be equal to the first-defined version of python in the test matrix.

.travis.yml
RobertRosca: Currently the test stage and the singularity image building stages are decoupled almost completely, the only relation between the two is that if the tests fail then the image won't build.

So the tests could be expanded on even more (e.g. additional compilers) without affecting the image build process, if that's what you mean.

.travis.yml
RobertRosca: I think something like this would be possible by just changing the line at the top from `os: linux` to:

```yml
os:
 - linux
 - osx
```

And then adding something like:

```yml
before_install:
 - if [ "$TRAVIS_OS_NAME" = "linux" ]; then export MINICONDA_OS="Linux-x86_64"  ; fi
 - if [ "$TRAVIS_OS_NAME" = "osx" ]; then export MINICONDA_OS="MacOSX-x86_64"   ; fi
```

That should then lead to  a 2x2 test matrix of tests running on both Linux and MacOSX, for two versions of python (3.7.5 and 3.6.9).

Singularity
RobertRosca: My bad, it was set like that from when I had some trouble setting up TBB in the image.

Updated the image definition with additional dependencies for TBB and removed those exports, now instead of using conda to install gcc/tbb dependencies it relies on CentOS Software Collections which are a bit more reliable to work with.

Could you check the updated version?

(note: I'm changing the way the install logs are named, so ignore the log naming/counting bit)

.travis.yml
zhujun98: I do something like the following in my other projects. 
```
matrix:
    - os: linux
      addons:
        apt:
          sources:
            - ubuntu-toolchain-r-test
          packages:
            - g++-7
    - os: osx
      osx_image: xcode9.3
      compiler: clang
```

And we may not want to increase the number of tests to 2x2 because of Python versions. I think having one test for Python 3.6.9 should be enough and the rest should only use Python 3.7.5.

I raised this question since the matrix is removed. I am just wondering whether it is possible to keep it, or you have better ways to realize all the tests (in and not in containers as well as different compiler/Python versions).

.travis.yml
zhujun98: Yes! I would like to expand the test stage! See my comment above.

.travis.yml
zhujun98: Using `PYTHON_VERSION` looks better to me.

Singularity
```diff
@@ -23,34 +27,51 @@ From: centos:7.6.1810
...
+  # logno=0
+  # alias logbase='logbase=$(printf $SINGULARITY_STAGE"%02d" $logno)'
+  # alias loginc='logno=$((logno+1))'
+  # ##  logno prints and increments the log number, 'A00', 'A01', etc...
+  # logno_func() {
+  #   echo $SINGULARITY_STAGE$(printf "%02d" $_logno)
+  #   _logno=$((_logno+1))
+  # }
+
+  # alias logno='$(printf $SINGULARITY_STAGE"%02d" $_logno); __l=$(printf $SINGULARITY_STAGE"%02d" $_logno); _logno=$((_logno+1))'
```
zhujun98: Do we want to keep the above commented code?

Singularity
zhujun98: I see. Indeed there is a bug in my CMake and I am too lazy to fix it: 

The runtime tbb library (`libtbb.so`) can be different from the compile time one. On the online cluster, it always dynamically links to the system `libtbb.so`. Maybe this is what troubled you.

The new version looks good to me. Thanks!

zhujun98: The above-mentioned bug has been fixed in #11

.travis.yml
ebadkamil: Then `extra-foam.def` please.

Singularity
```diff
@@ -0,0 +1,103 @@
...
+  cd EXtra-foam
+  git checkout $GIT_COMMIT_HASH
+  git log -n 1 | tee -a $logpath/$logno-extra-foam-clone.log
+  git submodule update --init | tee -a $logpath/$logno-extra-foam-submodules.log
+
+  pip install . | tee -a $logpath/$logno-extra-foam-pip-install.log
+
+  ##  Export environment to concretise
+  conda env export -n base -f /.singularity.d/logs/$logno-conda-env-export.log
+  pip freeze > /.singularity.d/logs/$logno-conda-env-export.log
```
ebadkamil: We can also have `%runscript` section for easy to run `./extra-foam.sif {detector}`.

zhujun98:
Can you rebase to the new dev branch? (We keep a linear history whenever possible)
and 
Squash all of your commits into one? (We try to make a clean history whenever possible)

Thanks!

RobertRosca:
My goal was to have a matrix which could take in: OS, GCC version, and Python version. Where only one test would run as a basic test, the full test suite would run over the entire matrix on a merge to master, and a job which builds and pushes a singularity image executes on tag/merge to master.

Writing a test like that was apparently beyond my travis YAML skills so after a lot of time wasted messing around with it I gave up and just wrote the test cases out individually. It's kind of ugly but at least it works.

I do have a few comments:

 - The tests run on GCC 6 and 7, however the docs say to install extra-foam using gcc from the omgarcia channel in conda, which only provides version 6.1.0, not version 7. This is the method used in the singularity image, so the image always have gcc 6.1.0, is this okay?
 - Pushing the singularity image to sylabs cloud requires an access token. These tokens expire every 30 days and have to be regenerated. Doing this once a month isn't a big deal but it should be documented somewhere.
 - Currently the singularity image is only built when merging to master or when a tag is made, do you want it to be built for other branches, e.g. dev?

zhujun98:
Thanks a lot for your effort! @RobertRosca 

> The tests run on GCC 6 and 7, however the docs say to install extra-foam using gcc from the omgarcia channel in conda, which only provides version 6.1.0, not version 7. This is the method used in the singularity image, so the image always have gcc 6.1.0, is this okay?
 
gcc 6.1.0 is used in the operational conda environment, so the image should also use it.

> Pushing the singularity image to sylabs cloud requires an access token. These tokens expire every 30 days and have to be regenerated. Doing this once a month isn't a big deal but it should be documented somewhere.

Yes, you can document in `docs/developer`. Maybe one should also document once the singularity image cannot be pushed to sylab, there is an easy way to disable it to prevent this kind of incident from blocking release.

> Currently the singularity image is only built when merging to master or when a tag is made, do you want it to be built for other branches, e.g. dev?

By the way, can you rebase it to the dev and merge all the commits into one? This is also some work but it makes the history of `EXtra-foam` clean. Thank you!
For now, I would only build the image when merging `dev` to `master`, i.e. before a release/a tag being made.

RobertRosca:
> Yes, you can document in docs/developer. Maybe one should also document once the singularity image cannot be pushed to sylab, there is an easy way to disable it to prevent this kind of incident from blocking release.

Added a short explanation on this to the docs, also set the singularity build push stage to be allowed to fail so it won't block the release.

> By the way, can you rebase it to the dev and merge all the commits into one?

Done, I think(?)

> For now, I would only build the image when merging dev to master, i.e. before a release/a tag being made.

Sure, left it as it was. Only builds on merge to master and on tag, so pulling the master image always gives newest version, but previous versions exist on the tag.

Just need to add some secret variables to travis now.

docs/developer.rst
```diff
@@ -59,6 +59,10 @@ Release **EXtra-foam**
 - Merge the above change into the `dev` branch;
 - Merge the `dev` branch into the `master` branch;
 - Tag the `master` branch.
+- "singularity build push" stage in CI may fail due to token error
+  - sylabs.io token expires every 30 days
+  - Failure is allowed so this won't prevent PR merger
+  - Organisation accounts do not exist yet - contact Robert to update token
```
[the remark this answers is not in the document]
RobertRosca: Yeah... only other option would be to make a separate account specifically to use for this. Far as I can tell organizations will be added 'soon', but they'll probably be an enterprise feature.

zhujun98:
I have no clue about the CI. It seems to be running on another working branch.

zhujun98:
I cloned your repo and the unittest works locally. Also the error message

```
self = <extra_foam.gui.tests.test_main_gui.TestMainGuiCtrl testMethod=testTrXasCtrl>

    def testTrXasCtrl(self):

>       from extra_foam.gui.special_suite_ctrl_widgets.trxas_ctrl_widget import (

            _DEFAULT_N_BINS, _DEFAULT_BIN_RANGE,

        )

E       ModuleNotFoundError: No module named 'extra_foam.gui.special_suite_ctrl_widgets'
```
is something from MR #89

RobertRosca:
I deleted my old fork, made a new one, and just copied my files over into it.

I'll open a new PR, if that doesn't work either then... ¯\_(ツ)_/¯

zhujun98:
I will close this one since #92 has already been merged!

Issue #18, opened 2020-01-07
Provide general interface for pixel-wise gain and offset correction


zhujun98:
vf = vi * gain + offset

We will have two options:

1. gain and/or offset are constant;
2. load pixel-wise gain and/or offset from files.

zhujun98:
@ebadkamil

PR #25, opened 2020-01-08
Gain and offset (constant)

zhujun98:
@ebadkamil 

This is the first step for #18, which defines the interface and apply gain and offset correction with a constant. Indeed, there is a parameter (`background`) in the old implementation which is actually the `offset`.


zhujun98:
@ebadkamil I have a question. Will the gain/offset correction and dark subtraction happen at the same time? or Only one of them will be used.

zhujun98:
It is ready to review now!

In the next PR, we will deal with loading gain/offset matrices from files.

zhujun98:
Thanks! Rebased and will merge it later.

Issue #20, opened 2020-01-07
Display reference image in ImageTool


zhujun98:
The reference image is obtained by clicking `Set reference` or `Load reference` button in `ImageTool`.

PR #27, opened 2020-01-09
Load gain and offset constants from files

zhujun98:
Follow-up PR for #18 

- Load numpy array from files and apply pixel-wise correction;
- Merge 'dark' functionality into gain/offset correction;
- Implement #20 
- Improve C++ code interface and the Python counter part. Users can use the fast implementations by `from extra_foam.algorithms import mask_image_data, correct_image_data, nanmean_image_data`, etc.

image


zhujun98:
Thanks!

PR #29, opened 2020-01-13
Improve redis server

zhujun98:
This is based on #28

### DIFF
--- a/extra_foam/config.py
+++ b/extra_foam/config.py
@@ -165,6 +165,7 @@ class _Config(dict):
         "REDIS_MAX_MEMORY_FRAC": 0.2,  # must <= 0.5
         # password to access the Redis server
         "REDIS_PASSWORD": "karaboFAI",  # FIXME
+        "REDIS_LOGFILE": osp.join(ROOT_PATH, "redis.log")
     }
 
     # system configurations which will appear in the config file so that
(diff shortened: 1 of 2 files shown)


extra_foam/services.py
```diff
@@ -137,6 +139,20 @@ def start_redis_server():
             logger.error(f"Failed to config the Redis server.\n" + repr(e))
             sys.exit(0)
 
+        # Increase the hard and soft limits for the redis client pubsub buffer
+        # to 512MB and 128MB, respectively.
+        cli_buffer_cfg = (client.config_get("client-output-buffer-limit")[
+            "client-output-buffer-limit"]).split()
+        assert len(cli_buffer_cfg) == 12
+        soft_limit = 128 * 1024 ** 2  # 128 MB
+        soft_second = 60  # in second
```
ebadkamil: what is `soft_second` limit here?

extra_foam/config.py
```diff
@@ -165,6 +165,7 @@ class _Config(dict):
         "REDIS_MAX_MEMORY_FRAC": 0.2,  # must <= 0.5
         # password to access the Redis server
         "REDIS_PASSWORD": "karaboFAI",  # FIXME
```
ebadkamil: some point of time change this to be consistent.

zhujun98: I think this is a good password, which has history in it. I have a `FIXME` here since I need a better way to make it confidential.

extra_foam/services.py
zhujun98: If the buffer size is above the `soft_limit` for `soft_second`, the client will be disconnected. For `hard_limit`, the client will be immediately disconnected.

zhujun98:
Rebased! Thanks!

PR #38, opened 2020-01-22
Source management

zhujun98:
This PR aims to greatly improve the data source management via the source tree:

1. Control required sources in the `DataSourceTree`;
2. Allow filtering by value for all non-detector data sources;
3. Resign the config code. Now each instrument will has its own config file, e.g. scs.config.yaml, fxe.config.yaml. All the instrument sources will be set up in the config file. As you have noticed, we replace JSON with YAML since the latter is more readable and allow comments;
4. Add `AdqDigitizer`;
5. Implement streaming raw (AGIPD, LPD) data from files and also  'confirmed' streaming raw (AGIPD, LPD) data online;


zhujun98:
Comments and suggestions are continuously welcome.

PR #40, opened 2020-01-31
Super fast LPD and DSSC assembling

zhujun98:
- Implement Fast assembling for LPD and DSSC;
- Also add an option to stack the above two detectors without assembling.


zhujun98:
Take your time to review before it is merged into master.

tmichela:
Interesting PR

- is it a simple portage of the python to c++ of do you have a different implementation?
- do you have some figures to show? How much faster is it over extra-geom?
- do you have tests that show extra-geom and this have the same result?
- any extra features that do not exist in extra-geom?

zhujun98:
Thanks for being interested :-)

 > is it a simple portage of the python to c++ of do you have a different implementation?

I adopted the methodology used in extra-geom, i.e., to find out the first pixel and the diagonal one of each tile first and then copy them to the assembled image. However, I did not create those helper classes like GeomFragment in extra-geom since I don't since they are necessary. In C++ code, there is only one class for each detector. For example, LPD_1MGeometry in C++ and LPD_1MGeometryFast in Python. The Python one has the same necessary interface as extra-geom, like `output_array_for_position_fast`, `position_all_modules` and `from_h5_file_and_quad_positions`.

 > do you have some figures to show? How much faster is it over extra-geom?

I will post the figures in a minute. On online cluster, it is 10 times faster. It should be noted that the C++ code does not optimize the memory copy which has been done greatly in numpy. So I expect there is still plenty of room to further boost the performance.

 > do you have tests that show extra-geom and this have the same result?

Good question. For DSSC the assembled sizes are the same but they differ by 1 pixel in LPD. I think there is some rounding issues that I have no time to figure it out now. But for now, they are not important.

 > any extra features that do not exist in extra-geom?

I do not have so many plot functions in extra-foam. But I have a `stack-only` button to conveniently stack all the modules without assembling, just for fun. The new implementation also offer me a convenient way to support more masking options, for example, mask the edges of the detectors. Of course, one can do it in extra-geom, but I care about the performance in extra-foam.

zhujun98:
Extra-foam
image

Extra-geom
image

zhujun98:
By the way, users can choose one of the assemblers as they like.

tmichela:
Thanks @zhujun98

tmichela:
If this is >10 times faster as the python version, I guess that can be interesting for many other tools to use this. Do you think it is possible to move this to extra-geom at some point? Or it's implementation is too specific to extra-foam?

tmichela:
If I understand well this is the equivalent of the "snapped" geometry, right? You are not doing any kind of interpolation?

@takluyver FYI

zhujun98:
> If this is >10 times faster as the python version, I guess that can be interesting for many other tools to use this. 

It depends on what kind of analysis users are interested. If what they need is averaging over a run, then can do the averaging before assembling and it does not matter how fast the assembling is. However, if they need to do analysis pulse-by-pulse, for example, azimuthal integration, or do a ROI sum filtering first, they will need to assemble the image pulse-by-pulse. In this case, reducing the assembling time from 100 ms to 10 ms changes the game.

> Do you think it is possible to move this to extra-geom at some point? Or it's implementation is too specific to extra-foam? 

It is more about the group software policy. It can be easily moved into extra-geom, but one needs a liitle bit effort to make a manylinux wheel to make it distributable.

zhujun98:
> If I understand well this is the equivalent of the "snapped" geometry, right? 
More or less.

> You are not doing any kind of interpolation?
No interpolation for now.

takluyver:
Thanks for the pointer. There's too much code for me to realistically read through, but:

- I'm not against having compiled code in EXtra-geom if there's a real benefit. It obviously means extra work for testing and packaging, but if the benefits justify that, there's no fundamental obstacle.
- I believe the limiting thing for assembling images (without interpolation) should be the memory copies. It's not obvious to me why doing memory copies through numpy should be 10x slower than doing them in C++ - so can we identify what makes the EXtra-geom assembly slow? Can we get closer to this kind of performance with just Python code?

zhujun98:
For operations which can be vectorized, like copy a chunk of memory, `numpy` is even faster than non-optimized `C/C++` code (e.g. assignment in a for loop). However, there are true threads in C++ which can increase the performance by brutal force. 

But `numpy` also supports threading operation to a certain extend. As far as I can tell, `EXtra-geom` could be 4 x faster than the current version if using Python's threads. However, as I said, the C++ counterpart can be even faster after optimization, for example, using SIMD, which already exists in `numpy`.

takluyver:
Thanks Jun. I've opened an issue to look at using threads: https://github.com/European-XFEL/EXtra-geom/issues/9 .

If EXtra-geom's implementation got 4x faster, your implementation would still be something like 2.5x faster, plus whatever more you can gain by optimising, correct? Do you know where that comes from? Is there unavoidable Python/numpy overhead? Or something else that EXtra-geom leaves on the table? Does the difference depend how much data is being assembled at once?

zhujun98:
> If EXtra-geom's implementation got 4x faster, your implementation would still be something like 2.5x faster, plus whatever more you can gain by optimising, correct?

Yes, I believe the C++ code can be at least 2 x faster. Besides SIMD, some for loop-copies can also be replaced by `memcpy`. 

> Do you know where that comes from? Is there unavoidable Python/numpy overhead?

From my experience,  `numpy` only supports multi-threading to some extend, which means the performance does not scale linearly or quasi-linearly as the number of threads increases. For example, I used to do `nanmean` in parallized numpy code and the maximum speed-up is about 4 x with 8 threads, if I remember correctly. When I further increased the number of threads, the performance even became worth. The consequence is that I had to hardcode the number of threads. For the assembling case, we can only know the exact result once it is implemented. 

> something else that EXtra-geom leaves on the table?

The most expensive part of `EXtra-geom` is the following piece of memory copy code. For AGIPD and LPD, it does a 16 by 16 for loop and it will have some non-negligible overhead. I would first try to parallize the for loop on modules to see the benefit.

```py
        for i, module in enumerate(self.modules):
            mod_data = data[..., i, :, :]
            tiles_data = self.geom.split_tiles(mod_data)
            for j, tile in enumerate(module):
                tile_data = tiles_data[j]
                # Offset by centre to make all coordinates positive
                y, x = tile.corner_idx + self.centre
                h, w = tile.pixel_dims
                out[..., y : y + h, x : x + w] = tile.transform(tile_data)
```

> Does the difference depend how much data is being assembled at once?

For this, one can only tell after the implementation. To be frankly, numpy's multi-threading operation is a black-box for me.

PR #44, opened 2020-02-03
Enhancement of statistics

zhujun98:
- Correlation and histogram plots are separated again;
- Allow automatically choosing bin range. In the worst case, the cost is less than 10 ms. Nevertheless, the user is encouraged to set up the bin range manually;
- Improve the visualization of heatmap.


zhujun98:
Take your time to review before it is merged into master.

PR #45, opened 2020-02-04
Improve pipeline behavior

zhujun98:
- Support multiple endpoints connection;
- Correlate data from the same/different endpoints with train ID;
- Allow select pipeline policy (wait or drop) via commandline. The default is wait since the data arrival speed is slower than the processing speed during online analysis;
- Replay Python's build in `queue.Queue` to speed up data transfer.

However, the current implementation is still crippled:
    1. All the connections must be alive;
    2. It is blocked, which means if one connection is fast and the other
       is slow, the overall performance is limited by the slow one.

Since this is not an indispensable feature for 0.8.0. I will conclude soon with some addition unittests.


zhujun98:
0.7.3 will be released this afternoon. This is an important MR and we will make use of this week to see its stability in the hutch.

PR #55, opened 2020-02-18
ROI histogram

zhujun98:
Feature requested by the detector people.

For pulse-resolved detectors. For train-resolved one, there will be a follow-up PR.

image


zhujun98:
Thanks!

PR #61, opened 2020-02-21
Improve masking behavior

zhujun98:
Return a boolean mask for each masking operation.


zhujun98:
Since MID and Jola have been using it for the whole day, I assume this major change in principle works.

PR #72, opened 2020-02-28
single module operation of a multi-module detector

zhujun98:


zhujun98:
Offline LGTM from @ebadkamil via phone call.

PR #74, opened 2020-03-01
Fixes #73

### LINKED ISSUE #73: Typo   'asssert_called_once'

scottwedge:


zhujun98:
Thanks a lot! I am curious how did you spot this bug in the test since `MagicMock` object does not raise `AttributeError` and that's why unittest does not complain.

PR #87, opened 2020-03-03
Implement save and load metadata.

zhujun98:


zhujun98:
We will have a powerful configurator next week :-)

image

PR #101, opened 2020-03-10
add_support_for_fast_adc

### LINKED ISSUE #100: Add support for FastADC
There is a channel "data.peaks" in this device, which is the same as "pulse_integral" in AdqDigitizer.

zhujun98:
Fix #100


zhujun98:
For now, some hack for test

PR #102, opened 2020-03-10
Implement AGIPD 1M geometry in C++

zhujun98:


zhujun98:
image

zhujun98:
Due to the conflict with #87, I will merge this one first. But the default geometry for AGIPD is EXtra-geom.

PR #105, opened 2020-03-12
0.8.1 release

zhujun98:


setup.py
```diff
@@ -31,9 +31,10 @@
 def find_version():
     with open(osp.join('extra_foam', '__init__.py')) as fp:
         for line in fp:
```
ebadkamil: Why not read entire file in a string ` f.read()`

`m = re.search(r'^__version__ = "(\d+\.\d+\.\d[a-z]*\d*)"', f.read(), re.M)`

zhujun98: The `re` search is line by line. `^__version__` is for the start of a line.

zhujun98: Resolved in offline discussion. Both solutions work.

zhujun98:
Thanks for the review!

PR #110, opened 2020-03-16
Mask tile edge

zhujun98:


zhujun98:
Finally, I catched the bug.

image

PR #111, opened 2020-03-16
Release 0.8.1

zhujun98:


zhujun98:
Thanks!

RobertRosca:
Not sure what's causing the failure exactly, or why allowed failures isn't allowing the merge to happen even when this fails.

I think that changing the jobs section to this might fix it:

```
jobs:
  allow_failures:
    - stage: singularity build push
  include:
    - stage: basic
      env: PYTHON_VERSION="3.7.5" COMPILER="gcc" GCCv="7"
    - stage: full linux
```

RobertRosca:
Cool, seems to allow failures alright now.

I'll do another patch pr with a fix for then cause of the error later today.

PR #112, opened 2020-03-16
Update .travis.yml

RobertRosca:
Actually allows singularity build stage to fail


zhujun98:
Thanks! Let's see whether it works.

PR #113, opened 2020-03-16
Merge pull request #111 from European-XFEL/dev

zhujun98:


zhujun98:
@RobertRosca why the singularity build was triggered when I merge `master` into `dev`?

zhujun98:
Also, there is the basic build with Python 3.7 and gcc 7.

zhujun98:
@RobertRosca The basic test should run with gcc6, which is the version used in operational environment. Can you also change this in your next PR? Thanks!

Issue #121, opened 2020-03-18
Upgrade EXtra-foam offline


zhujun98:
As the facility is in deep-shutdown status, people may want more offline features. @ebadkamil @RobertRosca you are welcome to work on that if you feel bored with other stuff. I can think of two points:

1. Improve the streamer;

2. Allow load the run, calculate the averaged data over the run;

3. Allow to visualize the data train-by-train.

In a word, we need a more generalized `loader`.

#122

zhujun98:
See also #123 #124

PR #122, opened 2020-03-19
Generalize file server

zhujun98:
As mentioned in #121, the goal is to provide a generalized streamer. What I would like to achieve in this PR are:

1. The streamer can be started by only typing `extra-foam-stream`. The detector name and port are not required (port is optional) and users can change them in the GUI;
2. We don't need the long info, which duplicates the data source information;
3. Separate pipeline and slow data source list widget;
4. Add future control widget placeholder, e.g. `mode`, which only has one option "continuous" for now;
5. Allow selecting property for each device which solves the slowdown issue when streaming detector data together with a single motor data.


zhujun98:
@ebadkamil Now I remember that you still own me the unittest :-)

zhujun98:
image

To be further improved in the coming MR.

PR #143, opened 2020-04-03
WIP: Use transformation of the corrected image as reference

### LINKED ISSUE #106: Generalize the design of reference tab in ImageTool
1. Add an option to make it a transformation of the original image, e.g. a Fourier transformation, a distortion.

This is a request from user run 5. See https://in.xfel.eu/redmine/issues/57328

2. The reference image and the original one do not necessarily to have the same shape. The shape check should be moved to the time when the reference image is used.

zhujun98:
Fix #106


zhujun98:
# implemented in #203

PR #146, opened 2020-04-06
Implement XAS-TIM in special suite

zhujun98:


zhujun98:
More test will be added later when #162 is implemented.

PR #168, opened 2020-04-22
Integrate client for karabo interface

### LINKED ISSUE #97: Communication with Karabo
During the training today, Steffen pointed out two important points:

1. Is it possible to guarantee that the data being analyzed and visualized online is indeed sent to the DAD and saved there?

2. The data flow scenario 3, which I describe as the final goal, is not ideal. This will  duplicate number of devices in each topic, with one bridge for each important detector. Steffen suggested to have a middlelayer device in Karabo which can receive request from EXtra-foam and organize the data. We will setup a meeting with him!

zhujun98:
Fix #97


zhujun98:
@tmichela Sorry for the long preparation for the integration, since I need enough special-suite apps to cover the various use cases. Also, it is much easier to test with a special suite app than with the main EXtra-foam instance.

I will summarize the feedback here in the coming days. @ebadkamil for info.

1. We need a way to switch between a `gate` and a traditional `bridge`;

tmichela:
Tha callback is there for the data to be immediately send out when received. So you don't have to loop on the `next` method until the next data is coming.

> Here the data does not need to be stored in a dictionary. I would like the data immediately be sent out when received.

it is sent out immediately with the callback. I want to keep internally a dict with all slow data, so if a client does not need callback on update he can just pick the current value of a property in the dict.

> Here, I prefer the data to be in the same format (data, meta) as the data sent out from the karabo bridge. Then the code outside the client does not need to be changed.

this is a bit weird for slow data, because it contains a single value...

note that this is just a simple prototype. it can change a lot.
If I understand you, you want the same interface as the karabo bridge, a `next` method, only you can register properties/channels to monitor?

I was also thinking of an asyncio based client, would that be something for extra-foam?

zhujun98:
Thanks for the explanation!

> If I understand you, you want the same interface as the karabo bridge, a next method, only you can register properties/channels to monitor?

In my opinion, for all the online analysis software, it should be able to work both with a normal karabo bridge and a karabo gate. The difference between these two scenarios is that the latter needs to ask for data first. So it would be nice if the receive part (here the `next` method) of these two scenarios are the same. Then the effort of adding support for karabo gate on a software which already supports karabo bridge is minimized. Of course, you can provide a more complicated karabo gate with more flexibilities. For example, each data source has its customized callback function.  But I would say a simple version is also required. And I prefer to have this simple version as a start.

> this is a bit weird for slow data, because it contains a single value...

First, fast data is just an array represented by a single string (property), which is the same as a slow data at this point.

Second, an online analysis software may or may not have the "matching" function. Unless you provide a matcher outside Karabo, like the one in EXtra-foam or EXtra-metro, we may want to consider the use case that a user software wants to use the karabo gate together with TrainMatcher. In this case, several data sources can be received together.

> I was also thinking of an asyncio based client, would that be something for extra-foam?

For now it is not the case. I am still overwhelmed with feature requests and have no time to touch the infrastructure. But I have been considering making workers "Actors" for a long time and then asyncio based client would be interesting.

tmichela:
> First, fast data is just an array represented by a single string (property), which is the same as a slow data at this point.

> Second, an online analysis software may or may not have the "matching" function. Unless you provide a matcher outside Karabo, like the one in EXtra-foam or EXtra-metro, we may want to consider the use case that a user software wants to use the karabo gate together with TrainMatcher. In this case, several data sources can be received together.

Sorry, that's not exactly what I meant. With slow data, I was referring to broker data (in opposition to pipeline data). Broker data is will always be a single key (source + property) and value (+timestamp/trainid). Pipeline data can have several keys/data (associated with a single timestamp/tid/sourcename).
If we use the trainMatcher, then the broker data become pipeline data (they are aggregated) and sent via pipeline (in karabo).

But now I see a bit more what you want, so I'll go a different route. I Probably will refactor slightly the current PipeToZeroMQ device to make it fit your requirement. that might be easier than making a new interface from scratch.

zhujun98:
> Sorry, that's not exactly what I meant. With slow data, I was referring to broker data (in opposition to pipeline data). Broker data is will always be a single key (source + property) and value (+timestamp/trainid). Pipeline data can have several keys/data (associated with a single timestamp/tid/sourcename).

I see. I am also not sure which is the better solution. It is just because for now, the way I proposed is more convenient for me to implement.

> But now I see a bit more what you want, so I'll go a different route. I Probably will refactor slightly the current PipeToZeroMQ device to make it fit your requirement. that might be easier than making a new interface from scratch.

I also just realized what exactly I need in the current stage :-) Let's keep it simple at the beginning. Otherwise, the users will turn their back to the new feature. Indeed it might not be very easy due to the 'complexity' of Karabo.

zhujun98:
@tmichela Here is a better place than Zoom.

How do you want to proceed? I tend to merge this PR. The new client in EXtra-foam special suite can only be activated by typing "--use_gate" explicitly. So it won't affect the traditional mode.

Now the problem is how to switch the `PipeToZeroMQ` device between an active (old) bridge and a passive (new) bridge.

tmichela:
I can make a new device class (`PipeToFoam` or the like :D) and you could run both devices at the same time for now? I suppose that in the long run you want to only use the 'gate' inteface.

There are also some other changes I want to do to this interface, at least: have individual sockets per channel connection (server side), so that if one source is faster it does not clogs the queue.
But that's not urgent I guess...

What was the status about feedback? I don't remember the conclusion with steffen... I think he mentioned using the scantool, but I guess extra-foam could send commands to it?

zhujun98:
> I can make a new device class (PipeToFoam or the like :D) and you could run both devices at the same time for now?

Sounds good to me! I tend to deploy the new "gate" on the online cluster and encourage the scientists to use it. At least, for apps like `CameraView` in special suite, the new "gate" should be able to work smoothly. Scientists should favor it because they do not need to configure the bridge to switch cameras/2D detectors. I will also encourage the SCS people to use the new XasTim with it.

> There are also some other changes I want to do to this interface, at least: have individual sockets per channel connection (server side), so that if one source is faster it does not clogs the queue.
But that's not urgent I guess...

I like it. Yes, it is not urgent since we still need to collect the first feedback from the operational envrionment. However, due to the current situation, this won't be quick.

> What was the status about feedback? 

Which feedback? You only have feedback from me :-)

> I don't remember the conclusion with steffen... I think he mentioned using the scantool, but I guess extra-foam could send commands to it?

Steffen suggested to try the tool (middleware?) that Istvan wrote first to interact with scantool. Istvan said he would invite us to a training that he would give to MID. Then everything stopped because of Cov19 ... Nevertheless, he said he is not against sending command directly from EXtra-foam to Scantool.

tmichela:
OK, what's a good timeplan for you? I can add the device tomorrow already (maybe) if you want it quicky

>  Nevertheless, he said he is not against sending command directly from EXtra-foam to Scantool.

Yep, that was what I meant with feedback, we should maybe discuss again about interfaces for it when you'll want to do this.

zhujun98:
No hurry! It would be nice to have it by the end of this week.

Regarding the discussion, we will need someone who is familiar with the scantool. Have you used it before? Moving a motor from A to B is a good starting point. Then we can explore more advanced functionalities in scantool, e.g. meshscan, etc.

extra_foam/pipeline/f_zmq.py
```diff
@@ -215,3 +220,189 @@ def __iter__(self):
...
+        self.connected = True
+        self._hb = Thread(target=self._heartbeat, daemon=True)
+        self._hb.start()  # start pinging server
+
+    def _heartbeat(self):
+        while self.connected:
+            sleep(10)
+            try:
+                self.ask({'request': 'ping'})
+            except zmq.error.ZMQError:
```
zhujun98, line 315: @tmichela As discussed here https://git.xfel.eu/gitlab/karaboDevices/PipeToZeroMQ/merge_requests/51#note_168347, it might be good to capture the exception and make a clean terminal. The `ping` message is simply used to keep the server alive so the client user does not need to know whether the ping succeeded or not.

zhujun98:
@tmichela I am about o merge this PR. Please let me know if you have any comments. The directly relevant code is in the class `QThreadKbClient`. Of course, comments on other parts are welcome.

extra_foam/pipeline/f_zmq.py
tmichela: That's right, although if the server does not reply, that probably means it's not working properly. I guess there is a better way to expose that to the client... But that probably helps in extra-foam if you see something like "server did not reply to ping, check what's going on there..."

tmichela:
Codewise this looks good to me

extra_foam/pipeline/f_zmq.py
zhujun98: Of course, it helps if there is some error message. The problem is that the `ping` is running in a daemon thread. It could be achieved by returning a state of the client in the `next` method.

PR #175, opened 2020-04-26
Implement generalized geometry for multi-module detectors

zhujun98:


zhujun98:
Finally, there is something less boring to do.

image

PR #279, opened 2020-09-23
WIP: Moving average across trains and intra-dark offset correction

philsmt:
This PR is not meant to be merged into master, but for any discussion related to a temporary implementation for requirements expressed in https://in.xfel.eu/redmine/issues/72646. These are needed for p2599 and p2471 in the beginning of October.

- The moving average uses the infrastructure in place for feature extraction and simply switches the data taken from the `ImageData` model in `ImageAnalysis` from `masked_mean` to `masked_mean_ma`. Said widget adds a checkbox to its context menus to toggle this behaviour. The control of the moving average window uses the existing metadata, which is also wired up to the existing image tool's input element (which has been a placeholder so far).
As `ImageAnalysis` is only used in the "Overview" so far, this has no effect on the visualizations in the "Gain/offset" tab. If necessary, the implementation could be moved to `ImageViewF` instead.

- The intradark offset correction is based on Loic's previous work here https://github.com/leguyader/EXtra-foam/commit/b903647621608d48a60e002ab70930073861d9d0. While the integration is not optimal, more refactoring would be needed to process data both with and without POI in `ImageProcessor` right now. A checkbox is added to the "Gain/offset" tab to toggle this new correction, which is mutually exclusive with any other correction. This is enforced for offset, but not gain currently.

* I've also cherry-picked a "fix" related to DSSC data values (https://github.com/leguyader/EXtra-foam/commit/b4afa99b73b1f0c05adac8881026da16b5129410). 

@ebadkamil

### DIFF
--- a/extra_foam/algorithms/imageproc_py.py
+++ b/extra_foam/algorithms/imageproc_py.py
@@ -12,7 +12,8 @@ import numpy as np
 from .imageproc import (
     nanmeanImageArray, movingAvgImageData,
     imageDataNanMask, maskImageDataNan, maskImageDataZero,
-    correctGain, correctOffset, correctGainOffset
+    correctGain, correctOffset, correctGainOffset,
+    correctIntraDarkOffset
 )
 
 
@@ -35,7 +36,7 @@ def nanmean_image_data(data, *, kept=None):
     return nanmeanImageArray(data, kept)
 
 
-def correct_image_data(data, *, gain=None, offset=None):
+def correct_image_data(data, *, gain=None, offset=None, intradark=False):
     """Apply gain and/or offset correct to image data.
 
     :param numpy.array data: image data, Shape = (y, x) or (indices, y, x)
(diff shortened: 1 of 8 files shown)


philsmt:
Available for testing on Maxwell now as `EXtra-foam/scs`

zhujun98:
Hi @philsmt, Thanks for working on the SCS features!

1. Regarding the implementation

> As ImageAnalysis is only used in the "Overview" so far, this has no effect on the visualizations in the "Gain/offset" tab. If necessary, the implementation could be moved to ImageViewF instead.

ImageViewF is the base class for all the rest ImageViews. It would confuse the users a lot to put a moving average checkbox there and it will not work. One could think of adding the moving average calculation in the GUI but it is an even worse idea which will immediately make EXtra-foam sluggish. 

Moreover, adding the moving average to a single image is not a good idea because it confuses users after they opened a lot of windows. And they will suffer from it by digesting a mixture of moving averaged and non-moving averaged images as well as the relevant analysis. 

You should have noticed that there is a disabled LineEdit for moving average since you enabled it. The moving average was there before but I removed it since it did not use the non-mean. But for DSSC it should be perfectly fine since there is no Nans from calibration. And for the other detectors it should be OK as long as the calibration does not end up with a lot of scattered Nan values across pulses.

2. Concerning the intradark offset, I agree the code needs refactoring.

> A checkbox is added to the "Gain/offset" tab to toggle this new correction, which is mutually exclusive with any other correction. 

I would use a `ComboBox` here as there could come more specific correction methods.

3. It would be nice to separate the two features. But you already said this PR is not meant for merge into the master. By the way, PRs of EXtra-foam used to be merged in to `dev` and only releases appear in the master branch.

zhujun98:
Since users would also like to apply any analysis on the averaged image, I applied a new implementation on #281. Now I remember that for detector like DSSC, we need to apply the moving average on the corrected data, which is a bit tricky.

JamesWrigley:
Closing this because the requested features were added in #285.

PR #281, opened 2020-09-28
Enable moving average of image for global analysis

### LINKED ISSUE #280: Image averager
In general, we would need simply an averager, in order to be able to see the signal-to-noise improving as we increase the number of trains averaged. **If this could become a general feature then we could apply to the averaged image, other operations already available in EXtra-Foam, like azimuthal integration for example.**

zhujun98:
Fix #280


zhujun98:
I need to think about it carefully as the image processing is complicated and error-prone. I want to blame Python as it is difficult to figure out which one is passed by reference.

Still, unittests are missing.

PR #294, opened 2021-03-25
Add initial support for saving and restoring sessions

JamesWrigley:
This has been requested because sometimes extra-foam will crash (particularly on
long-running instances) and the users will lose all settings they've set. These
can be very tedious and error-prone to redo, hence the desire to restore the
settings from the previous session.

Doing this for _all_ settings in extra-foam is a huge amount of work, so this
initial implementation only supports restoring the ROIs, which is one of the
more important settings.

(P.S., anyone else, feel free to add yourself as reviewer :) )

### DIFF
--- a/extra_foam/config.py
+++ b/extra_foam/config.py
@@ -8,6 +8,7 @@ Copyright (C) European X-Ray Free-Electron Laser Facility GmbH.
 All rights reserved.
 """
 from enum import IntEnum
+import os
 import os.path as osp
 import shutil
 from collections import abc, namedtuple
@@ -15,6 +16,8 @@ from collections import abc, namedtuple
 import yaml
 from yaml.scanner import ScannerError
 from yaml.parser import ParserError
+
+from PyQt5.QtCore import QObject, QSettings, pyqtSignal
 
 import numpy as np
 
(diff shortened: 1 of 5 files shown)


CammilleCC:
Cool and quick addition :D 

Is it possible to track the ROI visibility and lock states as well? This would help avoid going to the `ImageTool` window and toggling these back.

It would be great to have some tests as well.

extra_foam/config.py
```diff
@@ -26,6 +29,64 @@
...
+        self._settings = None
+        os.remove(self._current_session_path)
+        os.rename(self._last_session_path, self._current_session_path)
+        self._reset()
+
+        # Now that the new (old) settings are loaded, signal the widgets to
+        # restore themselves.
+        self.restore_session_sgn.emit()
+
+    def setValue(self, *args, **kwargs):
```
CammilleCC, line 74: Unless the `Session` single be expanded in the future, I believe it would be better for now to have the args as `setValue(self, key, value)` to reflect the `QSettings` signature.

CammilleCC: Same as for `value` and `contains`

extra_foam/gui/ctrl_widgets/roi_ctrl_widget.py
```diff
@@ -101,6 +101,10 @@ def initConnections(self):
     def setLabel(self, text):
         self._activate_cb.setText(text)
 
+    @property
+    def idx(self):
```
CammilleCC, line 105: `index` would be better (and should be alright, doesn't have a `QWidget` concurrence)

daviddoji:
After @CammilleCC review, nothing much to add.

It would be also great to have similar functionality for plots, as it also has been requested by many scientist. Sometimes during experiments, scientists know that a measurement is successful by looking at the plots. Reproducing the very same plots afterwards from h5 files could be not straigthforward for the non-experienced ones. 
Although this request might fit better in a different PR.

ebadkamil:
Settings can also be restored from here if they were saved.

Screenshot from 2021-03-26 12-14-52

This feature of loading settings may address the issue which this PR intends to. Or I am missing something? @zhujun98

zhujun98:
> Settings can also be restored from here if they were saved.
> 
> Screenshot from 2021-03-26 12-14-52
> 
> This feature of loading settings may address the issue which this PR intends to. Or I am missing something? @zhujun98

I have no idea what is happening here  :angry:

CammilleCC:
Thanks for the lead, guys! This is the feature that we're looking for :) 

The `Last Saved` will be helpful for the intermittent closing of the application, while the list of snapshots are nice for restore points.

Unfortunately, it was not evident for the users (and us :( ) that such feature exist.. we'll endorse this to the users.

ebadkamil:
> Unfortunately, it was not evident for the users (and us :( ) that such feature exist.. we'll endorse this to the users.

This feature was advertised during user trainings by @zhujun98 . documented [here](https://extra-foam.readthedocs.io/en/latest/main_gui.html#analysis-setup-manager) in case more information is needed

zhujun98:
Btw. I have never heard Extra-foam crashed before. FXE used to keep it running for months.

JamesWrigley:
Welp, closing this then :P

> Btw. I have never heard Extra-foam crashed before. FXE used to keep it running for months.

It was actually FXE that said that extra-foam was crashing, though I don't know any more details than that.

zhujun98:
Maybe some guy @ebadkamil wrote into the Redis database illegally :grin: 

To be serious, there could be two instances running (one forgot to close one of them), or the online cluster was in an unhealthy status, i.e. the memory was eaten up by a device, or the beamline scientist had no idea what he was doing :smile:

daviddoji:
@zhujun98 does the snapshot in the "Analysis setup manager" also store parameters defined by beamline scientist for a particular experiment, such as the ROIs?

zhujun98:
It stores all the values that you can type, check, tick, etc. in the GUI. So ROIs are included. Two things are not included:

1. The data source tree. I had no time for that before I left and it is indeed a bit tricky;
2. The specially suite which is not connected to Redis.

tmichela:
> Btw. I have never heard Extra-foam crashed before. FXE used to keep it running for months.

@JamesWrigley  also spotted a memory leak some time ago, don't know if that's related, but might :shrug:

zhujun98:
It would be nice to post an issue with the error message if anything suspicious happened. It could be difficult to reproduce it.

zhujun98:
If you have re-deployed it could also happen since the several new dependencies will even break the C++ code.

Issue #295, opened 2021-04-16
Recording reference image over trains


CammilleCC:
Hi all,

There is a request from SPB to record reference image over the trains, amongst others. I was wondering if the best location to put the mean calculation is in the `ImageProcessor` to be similar with the dark recording and probably the best place to do computation, or in the `ReferenceView` to be analogous with the current image as reference mechanism.

What do you guys think? Would be happy for any input :D

CammilleCC:
Solved with #296

PR #296, opened 2021-04-30
Record and save reference over multiple trains

CammilleCC:
Screenshot from 2021-04-30 17-57-50

This feature enables recording reference over multiple trains. This also includes saving the reference file on a specified directory. As the reference file is already saved in the temp folder, we just copy it to the input path.

It is requested by SPB and is mentioned on #295.

### DIFF
--- a/extra_foam/gui/ctrl_widgets/ref_image_ctrl_widget.py
+++ b/extra_foam/gui/ctrl_widgets/ref_image_ctrl_widget.py
@@ -7,7 +7,10 @@ Author: Jun Zhu <jun.zhu@xfel.eu>
 Copyright (C) European X-Ray Free-Electron Laser Facility GmbH.
 All rights reserved.
 """
-from PyQt5.QtWidgets import QGridLayout, QLineEdit, QPushButton
+from PyQt5.QtCore import pyqtSlot
+from PyQt5.QtWidgets import (
+    QGridLayout, QHBoxLayout, QLabel, QLineEdit, QPushButton, QSizePolicy,
+    QSpacerItem, QWidget)
 
 from ..ctrl_widgets import _AbstractCtrlWidget
 from ..gui_helpers import create_icon_button
@@ -26,6 +29,14 @@ class RefImageCtrlWidget(_AbstractCtrlWidget):
 
         self.set_current_btn = QPushButton("Set current as reference")
 
+        # Record
+        self.record_label = QLabel()
+        self.record_label.setFixedWidth(430)
+        self.record_btn = QPushButton("Record")
+        self.record_btn.setCheckable(True)
+        self.save_btn = QPushButton("Save")
+        self.save_btn.setDisabled(True)
+
         self._non_reconfigurable_widgets = [
             self.load_btn
         ]
@@ -43,11 +54,30 @@ class RefImageCtrlWidget(_AbstractCtrlWidget):
 
         layout.addWidget(self.set_current_btn, 1, 0)
 
+        # Add widgets for recording reference
+        spacer = QSpacerItem(0, 0, QSizePolicy.Expanding)
+        record_layout = QHBoxLayout()
+        record_layout.setContentsMargins(0, 0, 0, 0)
+        record_layout.addSpacerItem(spacer)
+        record_layout.addWidget(self.record_label)
+        record_layout.addWidget(self.record_btn)
+        record_layout.addWidget(self.save_btn)
+        widget = QWidget()
+        widget.setLayout(record_layout)
+        layout.addWidget(widget, 1, 1)
+
         self.setLayout(layout)
 
     def initConnections(self):
-        """Override."""
-        pass
+        self.record_btn.toggled.connect(self._enableButtonsOnRecord)
+
+    def onStart(self):
+        super().onStart()
+        self.record_btn.setEnabled(True)
+
+    def onStop(self):
+        super().onStop()
+        self.record_btn.setEnabled(False)
 
     def updateMetaData(self):
         """Override."""
@@ -56,3 +86,12 @@ class RefImageCtrlWidget(_AbstractCtrlWidget):
     def loadMetaData(self):
         """Override."""
         pass
+
+    @pyqtSlot(bool)
+    def _enableButtonsOnRecord(self, is_recording):
+        # Disable/enable buttons when (not) recording
+        self.save_btn.setDisabled(is_recording)
+        self.set_current_btn.setDisabled(is_recording)
+
+        # Update record button text
+        self.record_btn.setText("Stop" if is_recording else "Record")
--- a/extra_foam/gui/image_tool/reference_view.py
+++ b/extra_foam/gui/image_tool/reference_view.py
@@ -9,17 +9,22 @@ All rights reserved.
 """
 import os
 import os.path as osp
+import shutil
 
+import numpy as np
 from PyQt5.QtCore import pyqtSlot
 from PyQt5.QtWidgets import QFileDialog, QGridLayout
 
 from .base_view import _AbstractImageToolView, create_imagetool_view
 from ..ctrl_widgets import RefImageCtrlWidget
 from ..plot_widgets import ImageViewF
+from ...algorithms import movingAvgImageData
 from ...file_io import write_image
 from ...ipc import ReferencePub
 from ...logger import logger
 from ... import ROOT_PATH
+
+REFERENCE_FILE = osp.join(ROOT_PATH, "tmp", ".reference.npy")
 
 
 @create_imagetool_view(RefImageCtrlWidget)
@@ -39,6 +44,10 @@ class ReferenceView(_AbstractImageToolView):
 
         self._pub = ReferencePub()
 
+        self._is_recording = False
+        self._recorded_image = None
+        self._count = 0
+
         self.initUI()
         self.initConnections()
 
@@ -54,8 +63,11 @@ class ReferenceView(_AbstractImageToolView):
     def initConnections(self):
         """Override."""
         self._ctrl_widget.load_btn.clicked.connect(self._loadReference)
-        self._ctrl_widget.set_current_btn.clicked.connect(self._setReference)
+        self._ctrl_widget.set_current_btn.clicked.connect(
+            self._setCurrentAsReference)
         self._ctrl_widget.remove_btn.clicked.connect(self._removeReference)
+        self._ctrl_widget.record_btn.toggled.connect(self._recordReference)
+        self._ctrl_widget.save_btn.clicked.connect(self._saveReference)
 
     def updateF(self, data, auto_update):
         """Override."""
@@ -64,6 +76,20 @@ class ReferenceView(_AbstractImageToolView):
             # Removing and displaying of the currently displayed image
             # is deferred.
             self._reference.setImage(data.image.reference)
+
+        if self._is_recording:
+            self._count += 1
+            # Update cached image and count
+            image = self._corrected.image
+            if self._recorded_image is None:
+                self._recorded_image = image.copy()
+            else:
+                # Average without storing images
+                movingAvgImageData(self._recorded_image, image, self._count)
+
+            # Update status display
+            status = f"Recording..\t\tUsed trains: {self._count}"
+            self._ctrl_widget.record_label.setText(status)
 
     @pyqtSlot()
     def _loadReference(self):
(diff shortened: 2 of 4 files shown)


CammilleCC:
Huh, somehow the CI build has failed D: will look more into it, but any leads are greatly appreciated!

JamesWrigley:
The error is that CMake can't find `tbb_stddef.h`, which is weird since it certainly could when I last opened a PR (#294). Maybe a new version of TBB was released without that file?

EDIT: that can't be quite right, because it seems to build TBB automatically.

zhujun98:
Set TBB version number in the CI.  Don't use version 2021.1 and later.

extra_foam/gui/image_tool/reference_view.py
```diff
@@ -99,76 +103,87 @@ def _loadReference(self):
...
     def _removeReference(self):
         """Remove the reference image."""
-        self._pub.set("")
-        self._ctrl_widget.filepath_le.setText("")
+        self._set_reference("")
+        self._reference.clear()
 
     @pyqtSlot(bool)
     def _recordReference(self, is_recording):
+        """Records the reference image aby the"""
```
dallanto: I don't get this comment (typo? incomplete?) ... (maybe not of great significance anyway)

extra_foam/gui/image_tool/tests/test_image_tool.py
```diff
@@ -382,6 +382,23 @@ def testReferenceCtrlWidget(self):
...
+        for received_train in range(1, 6):
+            corrected._image = image * received_train
+            # mock view update as it contains the averaging logic
+            view.updateF(ProcessedData(1), False)
+        # stop recording
+        QTest.mouseClick(widget.record_btn, Qt.LeftButton)
+        # retrieve reference value and verify it is the average of the trains
+        updated, ref = proc._ref_sub.update()
+        self.assertTrue(updated)
+        np.testing.assert_array_equal(image * 3, ref)
```
dallanto: Excuse the naive question, but why do you assert for ``image * 3``, after receiving 5 trains?

ebadkamil:
```python
ref = (image + 2*image + 3*image + 4*image + 5*image ) /  5

```

dallanto: Thanks Ebad, I see, like 3 being the mean (expectation value) in the series 1-5, sure thing 😄

extra_foam/gui/image_tool/reference_view.py
```diff
@@ -76,26 +103,98 @@ def _loadReference(self):
...
+                # Update label
+                text = "The current image has been applied as reference."
+                self._ctrl_widget.record_label.setText(text)
+
+    def _write_temp_reference_file(self, image):
+        if image is None:
+            return
+
+        if not osp.exists(osp.dirname(REFERENCE_FILE)):
+            os.mkdir(osp.dirname(REFERENCE_FILE))
```
ebadkamil: just to save one line. we can get rid of `if` statement above by using. 
```python
os.makedirs(osp.dirname(REFERENCE_FILE), exist_ok=True)
```

extra_foam/gui/image_tool/tests/test_image_tool.py
CammilleCC: I agree that it was kinda confusing :P I changed testing a bit to be more descriptive

extra_foam/gui/image_tool/reference_view.py
CammilleCC: Thanks for the suggestion! Will take it :D

extra_foam/gui/ctrl_widgets/ref_image_ctrl_widget.py
```diff
@@ -43,11 +54,30 @@ def initUI(self):
...
+
         self.setLayout(layout)
 
     def initConnections(self):
-        """Override."""
-        pass
+        self.record_btn.toggled.connect(self._enable_buttons_on_record)
+
+    def onStart(self):
+        super(RefImageCtrlWidget, self).onStart()
```
JamesWrigley: Nitpicking, I think just `super()` will work in Python 3.

extra_foam/gui/image_tool/reference_view.py
```diff
@@ -21,6 +22,9 @@
 from ...logger import logger
 from ... import ROOT_PATH
 
+IMAGE_FILE_FILTER = "All supported files (*.tif *.npy)"
+REFERENCE_FILE = osp.join(ROOT_PATH, "tmp", ".reference.npy")
```
JamesWrigley, line 27: I think it makes more sense to keep the reference file visible, since `ROOT_PATH` is already a hidden directory.

extra_foam/gui/image_tool/reference_view.py
```diff
@@ -21,6 +22,9 @@
 from ...logger import logger
 from ... import ROOT_PATH
 
+IMAGE_FILE_FILTER = "All supported files (*.tif *.npy)"
```
JamesWrigley: This variable is never used.

extra_foam/gui/image_tool/reference_view.py
```diff
@@ -65,6 +76,22 @@ def updateF(self, data, auto_update):
             # is deferred.
             self._reference.setImage(data.image.reference)
 
+        if self._is_recording:
+            # Update cached image and count
+            image = self._corrected.image
+            if self._recorded_image is None:
+                self._recorded_image = image
+            else:
+                # Average without storing images
```
JamesWrigley, line 87: Is it possible to use the [`RawImageData`](https://github.com/European-XFEL/EXtra-foam/blob/dev/extra_foam/pipeline/data_model.py#L152) class here? It already supports keeping a moving average, and it uses the C++ implementation of the moving average if possible.

extra_foam/gui/image_tool/reference_view.py
```diff
@@ -76,26 +103,107 @@ def _loadReference(self):
...
+            self._write_temp_reference_file(image)
+            if self._check_reference_file():
+                self._set_reference(REFERENCE_FILE)
+                # Update label
+                text = "The current image has been applied as reference."
+                self._ctrl_widget.record_label.setText(text)
+
+    def _write_temp_reference_file(self, image):
+        """Save the input image as a reference file (.npy).
+           Path is $EXTRA_FOAM_TEMP_FOLDER/.reference.npy"""
```
JamesWrigley: Nitpicking, this could just point to the `REFERENCE_FILE` variable.

extra_foam/gui/image_tool/reference_view.py
```diff
@@ -76,26 +103,107 @@ def _loadReference(self):
...
+        image = self._corrected.image
+        if image is not None:
+            self._write_temp_reference_file(image)
+            if self._check_reference_file():
+                self._set_reference(REFERENCE_FILE)
+                # Update label
+                text = "The current image has been applied as reference."
+                self._ctrl_widget.record_label.setText(text)
+
+    def _write_temp_reference_file(self, image):
```
JamesWrigley: Nitpicking, I think we should stick to camelCase method names for consistency with the rest of the file. But on that note, it is a little weird how the GUI code uses camelCase and the rest uses snake case :P

extra_foam/gui/image_tool/reference_view.py
```diff
@@ -76,26 +103,107 @@ def _loadReference(self):
...
+            return
+
+        os.makedirs(osp.dirname(REFERENCE_FILE), exist_ok=True)
+
+        try:
+            write_image(REFERENCE_FILE, image)
+        except ValueError as e:
+            logger.error(str(e))
+            # Remove previous reference file
+            os.remove(REFERENCE_FILE)
```
JamesWrigley: This will throw another exception if `REFERENCE_FILE` doesn't exist, which _could_ happen if writing the image fails because e.g. the disk is out of space so the file is never created. But I feel that that's fairly unlikely, so I dunno if it's worth bothering about. And if the disk is out of space, the operator has bigger problems anyway :D

extra_foam/gui/image_tool/reference_view.py
```diff
@@ -76,26 +103,107 @@ def _loadReference(self):
...
+            # Remove existing reference file
+            self._removeReference()
+        elif self._recorded_image is not None:
+            # Save recording to a temp file upon finishing
+            self._write_temp_reference_file(self._recorded_image)
+            if self._check_reference_file():
+                self._set_reference(REFERENCE_FILE)
+
+            # Reset widget and variables
+            self._check_reference_file()
```
JamesWrigley: Is this second call necessary since `self._check_reference_file()` is already called in the if-statement above?

extra_foam/gui/image_tool/reference_view.py
```diff
@@ -76,26 +103,107 @@ def _loadReference(self):
...
+
+        # Validate filepath
+        if not filepath:
+            return
+        if not filepath.lower().endswith(".npy"):
+            filepath += ".npy"
+
+        # Copy reference file from tmp folder to desired destination
+        if not osp.exists(osp.dirname(filepath)):
+            os.mkdir(osp.dirname(filepath))
```
JamesWrigley: Could use the `exist_ok` argument again here.

JamesWrigley (changes requested):
Apologies for taking such a long time to review this :( Mostly LGTM, the majority of the comments are just nitpicks.

extra_foam/gui/ctrl_widgets/ref_image_ctrl_widget.py
CammilleCC: I got used to using `super` in that manner. Will change as advised to follow the rest of the code.

extra_foam/gui/image_tool/reference_view.py
CammilleCC: Changed, thanks for the suggestion :D

extra_foam/gui/image_tool/reference_view.py
CammilleCC: You're right, I was hoping that we can slowly deviate away from the camelCase. Changed to follow the rest of the code.

extra_foam/gui/image_tool/reference_view.py
CammilleCC: Good catch! I haven't realized that corner case. Added a preliminary check before removing :D

extra_foam/gui/image_tool/reference_view.py
CammilleCC: Actually, no. IIRC these were two different logic before I merged them into one XD Removed.

extra_foam/gui/image_tool/reference_view.py
CammilleCC: Gotcha

extra_foam/gui/image_tool/reference_view.py
CammilleCC: I tried and I realized that the `MovingAverageArray` classes are meant to be used with the `_BaseProcessor` variants. These are enabled with the `MetaProcessor` metaclass. Prolly it's not worth to reimplement things for this relatively small code change.

We could revisit again when there are more relevant changes :D

JamesWrigley: Ahhh I see. Maybe we could use `extra_foam.algorithms.movingAvgImageData()` directly then? That's implemented in C++.

extra_foam/gui/image_tool/reference_view.py
```diff
@@ -76,26 +102,106 @@ def _loadReference(self):
...
+            return
+
+        os.makedirs(osp.dirname(REFERENCE_FILE), exist_ok=True)
+
+        try:
+            write_image(REFERENCE_FILE, image)
+        except ValueError as e:
+            logger.error(str(e))
+            # Remove previous reference file
+            if os.path.exists(REFERENCE_FILE):
```
ebadkamil: just for consistency use `osp` since it's already imported.

extra_foam/gui/image_tool/reference_view.py
```diff
@@ -76,26 +102,106 @@ def _loadReference(self):
...
+    @pyqtSlot()
+    def _saveReference(self):
+        """'Save' the reference file on the user supplied path by conveniently
+            copying the existing file to the desired path."""
+        if not self._checkReferenceFile():
+            return
+
+        filepath = QFileDialog.getSaveFileName(
+            caption="Save image",
+            directory=osp.expanduser("~"),
```
ebadkamil, line 190: may be a good idea to have `last_saved_location` from the filepath. and the next time if user wants to save again, use that as the starting directory in the FileDialog.

ebadkamil:
LGTM.

ebadkamil (approved):

extra_foam/gui/image_tool/reference_view.py
```diff
@@ -76,26 +102,106 @@ def _loadReference(self):
...
+            self._ctrl_widget.record_label.setText(text)
+
+    @pyqtSlot()
+    def _saveReference(self):
+        """'Save' the reference file on the user supplied path by conveniently
+            copying the existing file to the desired path."""
+        if not self._checkReferenceFile():
+            return
+
+        filepath = QFileDialog.getSaveFileName(
```
ebadkamil, line 188: `self` as a first argument is missing? or not required

extra_foam/gui/image_tool/reference_view.py
CammilleCC: I'd like that idea :D prolly it would be better to implement it in a more general manner. Would think more about this, perhaps on the next MRs.

extra_foam/gui/image_tool/reference_view.py
CammilleCC: Hmm, from what I've seen with the other file dialogs, there's no passed parent. Not sure if it's by design, but I ought to leave it as is.

extra_foam/gui/image_tool/reference_view.py
CammilleCC: Whoops, my automatism manifests. Changed.

extra_foam/gui/image_tool/reference_view.py
CammilleCC: Done! Thanks for the suggestion :D

extra_foam/gui/image_tool/reference_view.py
```diff
@@ -65,6 +77,21 @@ def updateF(self, data, auto_update):
...
             self._reference.setImage(data.image.reference)
 
+        if self._is_recording:
+            # Update cached image and count
+            image = self._corrected.image.astype(np.float32)
+            if self._recorded_image is None:
+                self._recorded_image = image.astype(np.float32)
+            else:
+                # Average without storing images
+                num_images = self._count + 1
```
JamesWrigley: `num_images` is unnecessary if you increment `self._count` at the beginning of the if-statement.

extra_foam/gui/image_tool/reference_view.py
```diff
@@ -65,6 +77,21 @@ def updateF(self, data, auto_update):
             # is deferred.
             self._reference.setImage(data.image.reference)
 
+        if self._is_recording:
+            # Update cached image and count
+            image = self._corrected.image.astype(np.float32)
```
JamesWrigley: Why is it necessary to explicitly convert to `np.float32`?

extra_foam/gui/image_tool/reference_view.py
ebadkamil: yes makes sense. Thanks

JamesWrigley (changes requested):

JamesWrigley:
Oh, and maybe the docs should be updated?

extra_foam/gui/image_tool/reference_view.py
ebadkamil: Good point. This is indeed an expensive op. Do it only if necessary. Most likely the image here will be the assembled one and that should be of `dtype` `np.float32`

extra_foam/gui/image_tool/reference_view.py
CammilleCC: I was thinking of only adding the count after the operations are finished, but I guess it's unlikely that the calculations would result into error XD will change.

extra_foam/gui/image_tool/reference_view.py
CammilleCC: Whoops, it's because I added tests of `int32` arrays. I forgot that the assembled images results to `np.float32`, so conversion shouldn't matter.

JamesWrigley (approved):
LGTM! Squash and merge :)

PR #299, opened 2021-06-04
Minor fixes/utils

JamesWrigley:
Two changes:
- Fix all deprecation warnings from Numpy 1.20
- Add a context manager to time blocks of code. I found this very useful to quickly time specific blocks.

### DIFF
--- a/extra_foam/utils.py
+++ b/extra_foam/utils.py
@@ -42,6 +42,36 @@ def profiler(info, *, process_time=False):
     return wrap
 
 
+
+class BlockTimer:
+    """A context manager for measuring the execution time of a code block
+
+    For example::
+
+        >>> with BlockTimer("foo"):
+        ...     time.sleep(1)
+        ...
+        Execution of foo: 1.001s
+    """
+    def __init__(self, label="block", enabled=True):
+        """Create the timer object
+
+        :param str label: A name to identify the block being timed.
+        :param bool enabled: Whether or not to enable this timer.
+        """
+        self._label = label
+        self._enabled = enabled
+
+    def __enter__(self):
+        self.start = time.perf_counter()
+        return self
+
+    def __exit__(self, *_):
+        duration = time.perf_counter() - self.start
+        if self._enabled:
+            logger.info(f"Execution of {self._label}: {duration:.4f}s")
+
+
 _NOT_FOUND = object()
 
 
--- a/extra_foam/algorithms/imageproc_py.py
+++ b/extra_foam/algorithms/imageproc_py.py
@@ -7,8 +7,6 @@ Author: Jun Zhu <jun.zhu@xfel.eu>, Ebad Kamil <ebad.kamil@xfel.eu>
 Copyright (C) European X-Ray Free-Electron Laser Facility GmbH.
 All rights reserved.
 """
-import numpy as np
-
 from .imageproc import (
     nanmeanImageArray, movingAvgImageData,
     imageDataNanMask, maskImageDataNan, maskImageDataZero,
@@ -102,7 +100,7 @@ def mask_image_data(arr, *,
         if arr.ndim == 3:
             raise ValueError("'arr' must be 2D when 'out' is specified!")
 
-        if out.dtype != np.bool:
+        if out.dtype != bool:
             raise ValueError("Type of 'out' must be bool!")
 
         if image_mask is None:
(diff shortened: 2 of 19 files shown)


extra_foam/utils.py
```diff
@@ -42,6 +42,31 @@ def timed_f(*args, **kwargs):
...
+        self._enabled = enabled
+
+    def __enter__(self):
+        self.start = time.time()
+        return self
+
+    def __exit__(self, *_):
+        duration = time.time() - self.start
+        if self._enabled:
+            print(f"Execution of {self._label}: {duration:.3f}s")
```
CammilleCC: I am kinda concerned with the resolution. While this suffice for big and slow blocks, one might find it useful to measure quick blocks. Should you like to extend it, perhaps `time.perf_counter` might be more fitting.

CammilleCC:
Just a small comment, otherwise LGTM :D 

I should update my `numpy` as well, haven't seen these warnings XD

extra_foam/utils.py
JamesWrigley: Good point, fixed in aa779423.

extra_foam/utils.py
```diff
@@ -42,6 +42,31 @@ def timed_f(*args, **kwargs):
...
+        self._enabled = enabled
+
+    def __enter__(self):
+        self.start = time.perf_counter()
+        return self
+
+    def __exit__(self, *_):
+        duration = time.perf_counter() - self.start
+        if self._enabled:
+            print(f"Execution of {self._label}: {duration:.4f}s")
```
tmichela: `logger.info`?

JamesWrigley: Oooo yes that's better, fixed in 175daf75.

CammilleCC: Good catch! I had the impression that this is an internal timer. I agree that it is better to use `logger.info` if we'd like to expose it outside. We can also use `logger.debug` similar with the the `utils.profiler`.

extra_foam/utils.py
```diff
@@ -42,6 +42,31 @@ def timed_f(*args, **kwargs):
...
+    """
+    def __init__(self, label="block", enabled=True):
+        self._label = label
+        self._enabled = enabled
+
+    def __enter__(self):
+        self.start = time.perf_counter()
+        return self
+
+    def __exit__(self, *_):
```
tmichela, line 69: nitpicking: we usually use `*args` and `**kwargs` for arguments and keyword arguments in python.
You could also use the positional arguments, even if you don't use them, ` exception_type, exception_value, traceback`

JamesWrigley: Hmm, I don't really like that style, to me it's a bit weird to take named arguments if they're never used. If they were used then I would indeed have used `*args` or `exception_type, ...`.

extra_foam/utils.py
```diff
@@ -42,6 +42,31 @@ def timed_f(*args, **kwargs):
...
+    """A context manager for measuring the execution time of a code block
+
+    For example::
+
+        >>> with BlockTimer("foo"):
+        ...     time.sleep(1)
+        ...
+        Execution of foo: 1.001s
+    """
+    def __init__(self, label="block", enabled=True):
```
tmichela, line 56: You could add a quick description of the arguments, even if they are obvious

JamesWrigley: True, added in 3ddcf067.

extra_foam/utils.py
tmichela: as said, it's just nitpicking about the common practice in Python, not forcing you to change it

tmichela (approved):

extra_foam/utils.py
```diff
@@ -42,6 +42,36 @@ def timed_f(*args, **kwargs):
...
+        self._enabled = enabled
+
+    def __enter__(self):
+        self.start = time.perf_counter()
+        return self
+
+    def __exit__(self, *_):
+        duration = time.perf_counter() - self.start
+        if self._enabled:
+            logger.info(f"Execution of {self._label}: {duration:.4f}s")
```
ebadkamil, line 72: `logger.debug` may be?

JamesWrigley: I think info is better because this class is only meant to be used for one-offs during development, and in that situation it's easier to not have to remember to run extra-foam in debug mode. Plus, there's a lot of other debug output which kinda clutters the terminal and makes it harder to see the line you're looking for.

PR #300, opened 2021-06-04
Implement photon binning

JamesWrigley:
This was part of MID's moonshot project, it helps to get rid of noise by binning the raw data by how many photons hit each pixel.

### DIFF
--- a/extra_foam/algorithms/imageproc_py.py
+++ b/extra_foam/algorithms/imageproc_py.py
@@ -7,12 +7,26 @@ Author: Jun Zhu <jun.zhu@xfel.eu>, Ebad Kamil <ebad.kamil@xfel.eu>
 Copyright (C) European X-Ray Free-Electron Laser Facility GmbH.
 All rights reserved.
 """
+import numpy as np
+
 from .imageproc import (
     nanmeanImageArray, movingAvgImageData,
     imageDataNanMask, maskImageDataNan, maskImageDataZero,
-    correctGain, correctOffset, correctDsscOffset, correctGainOffset
+    correctGain, correctOffset, correctDsscOffset, correctGainOffset,
+    binPhotons
 )
 
+
+def bin_photons(data, adu_count, out=None):
+    if data.ndim not in [2, 3]:
+        raise ValueError("Only 2D and 3D arrays are supported")
+
+    if out is None:
+        out = np.empty_like(data)
+
+    binPhotons(data, adu_count, out)
+
+    return out
 
 def nanmean_image_data(data, *, kept=None):
     """Compute nanmean of an array of images of a tuple/list of two images.
--- a/extra_foam/algorithms/tests/test_image_proc.py
+++ b/extra_foam/algorithms/tests/test_image_proc.py
@@ -5,11 +5,46 @@ import pytest
 import numpy as np
 
 from extra_foam.algorithms import (
-    correct_image_data, mask_image_data, movingAvgImageData, nanmean_image_data
+    correct_image_data, mask_image_data, movingAvgImageData, nanmean_image_data, bin_photons
 )
 
 
 class TestImageProc(unittest.TestCase):
+    def testPhotonBinning(self):
+        arr1d = np.full((2,), 100, dtype=float)
+        arr2d = np.full((2, 2), 100, dtype=float)
+        arr3d = np.full((2, 2, 2), 100, dtype=float)
+
+        # Test invalid inputs
+        with self.assertRaises(ValueError):
+            bin_photons(arr1d, 2)
+
+        with self.assertRaises(TypeError):
+            bin_photons(arr2d, 2, out=arr1d)
+
+        # Ensure that the out parameter is respected
+        arr3d_copy = arr3d.copy()
+        result = bin_photons(arr3d, 2)
+        np.testing.assert_array_equal(arr3d, arr3d_copy)
+
+        result = bin_photons(arr3d, 2, out=arr3d)
+        self.assertIs(arr3d, result)
+        self.assertFalse(np.array_equal(arr3d, arr3d_copy))
+
+        # Test for contiguous arrays
+        data = np.array([[-2, -1, 5, 0], [1.5, 2.5, 3, np.nan]])
+        expected_results = np.array([[0, 0, 3, 0], [1, 1, 2, np.nan]])
+        self.assertTrue(data.data.c_contiguous)
+        self.assertTrue(expected_results.data.c_contiguous)
+        np.testing.assert_array_equal(bin_photons(data, 2), expected_results)
+
+        # Test for non-contiguous arrays
+        data = data.swapaxes(0, 1)
+        expected_results = expected_results.swapaxes(0, 1)
+        self.assertFalse(data.data.c_contiguous)
+        self.assertFalse(expected_results.data.c_contiguous)
+        np.testing.assert_array_equal(bin_photons(data, 2), expected_results)
+
     def testNanmeanImageData(self):
         arr1d = np.ones(2, dtype=np.float32)
         arr2d = np.ones((2, 2), dtype=np.float32)
--- a/src/extra_foam/include/f_utilities.hpp
+++ b/src/extra_foam/include/f_utilities.hpp
@@ -13,9 +13,15 @@
 
 #include <string>
 #include <algorithm>
+#include <functional>
 
 #include "xtensor/xadapt.hpp"
 #include "xtensor/xio.hpp"
+
+#if defined(FOAM_USE_TBB)
+#include "tbb/blocked_range2d.h"
+#include "tbb/blocked_range3d.h"
+#endif
 
 
 namespace foam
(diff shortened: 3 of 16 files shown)


jo-moeller:
I learned during the beamtime, that the naming wasn't maybe that optimal. As the value that we call threshold is actually not the real threshold in the calculations, some people got confused by that.
Maybe replacing the name threshold with "ADUs/photon" or something like that?

JamesWrigley:
Ah I see, replaced it with `ADUs/photon` in 185ab809.

JamesWrigley:
I just pushed a rather large refactor. The main API change is that `binPhotons()` now takes an `out` parameter to write to.

This required changing the implementation quite a bit, which unfortunately is less flexible. The previous implementation simply iterated over the underlying buffer of the `pytensor`, and this works fine as long as only one `pytensor` is being operated on. I discovered through working on a separate thing that just iterating over the buffer is unreliable because the memory layout of the `ndarray` may not match the logical layout. For example, if a 2D `ndarray` is created with a [`NPY_ARRAY_C_CONTIGUOUS`](https://numpy.org/doc/stable/reference/c-api/array.html#c.NPY_ARRAY_C_CONTIGUOUS) layout and then `np.moveaxis()` is called on it, it will no longer be `NPY_ARRAY_C_CONTIGUOUS`. A very obvious bug in hindsight, but wonderfully tricky to debug :P

So if a C++ function is operating on two `pytensor`'s then the only safe way to iterate over the buffers is to ensure that both are contiguous and their layouts match. Otherwise the element at index `i` in one array will not match the logical location of the element at `i` in the other array and the output will be garbage. IMHO this is not a good design choice because it requires the user to ensure the contiguity of their `ndarray`'s, and this could require making a copy of the `ndarray` to have a contiguous layout, which might easily be slower than calling the C++ function. Hence the refactor in 61908013 to use [array indexing](https://xtensor.readthedocs.io/en/latest/indices.html#operators-array-index).

As I mentioned, this does make the C++ function less flexible because the code depends on the rank of the `pytensor`. I added a couple of helpers in 9b6e2d5d to simplify this. These required C++17 so I also rebased the branch onto #311 (which should be merged first).

JamesWrigley:
> I just pushed a rather large refactor. The main API change is that `binPhotons()` now takes an `out` parameter to write to.
> 
> This required changing the implementation quite a bit, which unfortunately is less flexible. The previous implementation simply iterated over the underlying buffer of the `pytensor`, and this works fine as long as only one `pytensor` is being operated on. I discovered through working on a separate thing that just iterating over the buffer is unreliable because the memory layout of the `ndarray` may not match the logical layout. For example, if a 2D `ndarray` is created with a [`NPY_ARRAY_C_CONTIGUOUS`](https://numpy.org/doc/stable/reference/c-api/array.html#c.NPY_ARRAY_C_CONTIGUOUS) layout and then `np.moveaxis()` is called on it, it will no longer be `NPY_ARRAY_C_CONTIGUOUS`. A very obvious bug in hindsight, but wonderfully tricky to debug :P
> 
> So if a C++ function is operating on two `pytensor`'s then the only safe way to iterate over the buffers is to ensure that both are contiguous and their layouts match. Otherwise the element at index `i` in one array will not match the logical location of the element at `i` in the other array and the output will be garbage. IMHO this is not a good design choice because it requires the user to ensure the contiguity of their `ndarray`'s, and this could require making a copy of the `ndarray` to have a contiguous layout, which might easily be slower than calling the C++ function. Hence the refactor in [6190801](https://github.com/European-XFEL/EXtra-foam/commit/61908013caf806c10770449d6869638b7c785d16) to use [array indexing](https://xtensor.readthedocs.io/en/latest/indices.html#operators-array-index).
> 
> As I mentioned, this does make the C++ function less flexible because the code depends on the rank of the `pytensor`. I added a couple of helpers in [9b6e2d5](https://github.com/European-XFEL/EXtra-foam/commit/9b6e2d5d2954800bd07c83c16f8e6c77e6402a59) to simplify this. These required C++17 so I also rebased the branch onto #311 (which should be merged first).

I've been re-thinking this, and I'm no longer sure it's the best way forward. The reason I refactored it was because I wanted to use the function in a special suite, with data that had been assembled with [`stack_detector_data`](https://extra-data.readthedocs.io/en/latest/agipd_lpd_data.html#extra_data.stack_detector_data) (which rearranges axes at some point and thus makes the array non-contiguous). However, in hindsight I think that whole thing could be done much more easily a different way, which removes the original need for this refactor.

Additionally, the C++ code here is... complicated..., and if we continue supporting non-contiguous layouts like this then all future functions we implement will also be as complicated (in addition to the existing C++ code). So, I now lean towards making a hard requirement that the input arrays have contiguous layouts, and throwing an error if they don't. I think this will make the code simpler, and also slightly faster in some cases. What do y'all think?

extra_foam/algorithms/tests/test_image_proc.py
```diff
@@ -5,11 +5,32 @@
...
+            bin_photons(arr2d, 2, out=arr1d)
+
+        # Ensure that the out parameter is respected
+        arr3d_copy = arr3d.copy()
+        result = bin_photons(arr3d, 2)
+        np.testing.assert_array_equal(arr3d, arr3d_copy)
+
+        result = bin_photons(arr3d, 2, out=arr3d)
+        self.assertIs(arr3d, result)
+        self.assertFalse(np.array_equal(arr3d, arr3d_copy))
```
CammilleCC, line 32: Thanks for adding tests! :D

While there is a C++ test for the actual binning logic, it would be also great to test the Python output here for completeness.

extra_foam/algorithms/imageproc_py.py
```diff
@@ -7,13 +7,28 @@
...
 
+def bin_photons(data, adu_count, out=None):
+    rank = len(data.shape)
+    if rank != 2 and rank != 3:
+        raise ValueError("Only 2D and 3D arrays are supported")
+
+    if out is None:
+        out = np.empty_like(data)
+
+    binPhotons(data, out, adu_count)
```
CammilleCC: Nitpick: I would suggest following the signature as observed from the other functions: `f(arr, *args, out)`. Also, using the same array when `out` is not specified.

src/extra_foam/include/f_utilities.hpp
```diff
@@ -54,6 +60,49 @@ inline void checkEven(T a, std::string&& header)
...
+#else
+      for (size_t i = 0; i < shape[0]; ++i) {
+        for (size_t j = 0; j < shape[1]; ++j) {
+          for (size_t k = 0; k < shape[2]; ++k) {
+#endif
+            functor(i, j, k);
+          }
+        }
+      }
+#if defined(FOAM_USE_TBB)
```
CammilleCC, line 101: Question: why do we ask `defined(FOAM_USE_TBB)` here again?

extra_foam/algorithms/imageproc_py.py
```diff
@@ -7,13 +7,28 @@
...
     imageDataNanMask, maskImageDataNan, maskImageDataZero,
-    correctGain, correctOffset, correctDsscOffset, correctGainOffset
+    correctGain, correctOffset, correctDsscOffset, correctGainOffset,
+    binPhotons
 )
 
 
+def bin_photons(data, adu_count, out=None):
+    rank = len(data.shape)
+    if rank != 2 and rank != 3:
```
CammilleCC: Nitpick: you could use `data.ndim` here, and `rank not in (2, 3)` prolly is more descriptive.

CammilleCC (commented):
Thanks for your work, I've tested it and it works great! :D 
I only have a few questions, but in general LGTM

Mmm, I would incline supporting non-contiguous arrays. I can't say much about the C++ utils that you've added, but seems that it works seamlessly. Could you also add a test for both non- and contiguous arrays just in case we go forward differently in the future?

extra_foam/algorithms/tests/test_image_proc.py
JamesWrigley: Added in 34f83680. Good thing you mentioned it, I found a wee bug related to NaN handling while adding the test :)

extra_foam/algorithms/imageproc_py.py
JamesWrigley: Changed the signature in 34f83680. I didn't default to modifying the input array in-place though, that doesn't match the behaviour of numpy/scipy so I think it'd be a little confusing to do it here.

src/extra_foam/include/f_utilities.hpp
JamesWrigley: It's to add the closing brace and parenthesis to match the call to `tbb::parallel_for()`. The single-threaded implementation is somewhat awkwardly nested inside the parallel implementation, so that's why this needs to be checked again at the end.

extra_foam/algorithms/imageproc_py.py
JamesWrigley: Fixed in 34f83680.

JamesWrigley:
Oh wait, forgot to add a test for [non-]contiguous arrays.

JamesWrigley:
Alrighty, added in c4b59c37.

CammilleCC (approved):
LGTM :D

PR #302, opened 2021-06-09
Implement curve-fitting for the azimuthal integration

JamesWrigley:
This was part of MID's moonshot project, it allows the user to curve fit the I(q).

image


CammilleCC (commented):
LGTM, cool addition!

CammilleCC (approved):

PR #303, opened 2021-06-11
Add extra FOMs for azimuthal integration

JamesWrigley:
This was part of MID's moonshot project, it implements some more azimuthal integration (AI) FOMs:
- Peak _I(q)_
- _q_ of the peak
- The _q_ of the center of mass of the _I(q)_

Unfortunately this breaks the abstraction of 'only one FOM per analysis type', which leads to some rather dubious code in e.g. `base_processor.py`. We could just re-use the `.fom` slot of `data_model.AzimuthalIntegrationData` for whichever FOM is currently selected, but that means that it wouldn't be possible to use multiple AI FOMs at once. I think that would be quite confusing for users; plus, using multiple AI FOMs simultaneously could be useful.

image

### DIFF
--- a/extra_foam/gui/image_tool/azimuthal_integ_1d_view.py
+++ b/extra_foam/gui/image_tool/azimuthal_integ_1d_view.py
@@ -61,6 +61,11 @@ class AzimuthalInteg1dPlot(PlotWidgetF):
 
     def setFitted(self, x, y):
         self._fitted.setData(x, y)
+        com_x, com_y = ai.center_of_mass
+        if com_x == np.nan or com_y == np.nan:
+            self._center_of_mass.setData([], [])
+        else:
+            self._center_of_mass.setData([com_x], [com_y])
 
 
 @create_imagetool_view(AzimuthalIntegCtrlWidget)
--- a/extra_foam/gui/image_tool/azimuthal_integ_1d_view.py
+++ b/extra_foam/gui/image_tool/azimuthal_integ_1d_view.py
@@ -7,6 +7,8 @@ Author: Jun Zhu <jun.zhu@xfel.eu>
 Copyright (C) European X-Ray Free-Electron Laser Facility GmbH.
 All rights reserved.
 """
+import numpy as np
+
 from PyQt5.QtCore import pyqtSlot
 from PyQt5.QtWidgets import QVBoxLayout, QSplitter, QTabWidget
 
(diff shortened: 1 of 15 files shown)


extra_foam/gui/image_tool/azimuthal_integ_1d_view.py
```diff
@@ -54,6 +56,12 @@ def updateF(self, data):
             self._peaks.setData(momentum[peaks], intensity[peaks])
             self.setAnnotationList(momentum[peaks], intensity[peaks])
 
+        com_x, com_y = ai.center_of_mass
+        if com_x is np.nan or com_y == np.nan:
```
CammilleCC: `com_x == np.nan`

extra_foam/pipeline/data_model.py
```diff
@@ -188,12 +188,15 @@ def __init__(self):
 
 class AzimuthalIntegrationData(DataItem):
     """Azimuthal integration data item."""
-    __slots__ = ['x', 'y', 'fom', 'q_map', 'peaks']
+    __slots__ = ['q_map', 'peaks', 'max_peak', 'max_peak_q', 'center_of_mass']
```
CammilleCC, line 191: We're still using `ai.x`, `ai.y`, and `ai.fom`, so I guess these shouldn't be removed

extra_foam/pipeline/processors/binning.py
```diff
@@ -368,7 +338,7 @@ def _update_data_point(self, processed, raw):
         if err:
             raise ProcessingError(err)
 
-        fom, vfom, vfom_x = ret.fom, ret.y, ret.x
+        fom, vfom, vfom_x = fom, ret.y, ret.x
```
CammilleCC: Remove `fom`?

CammilleCC (commented):
Great work of the refactoring and the `_FomProcessor` wrapper! :) In general it looks good aside from some code concerns.

extra_foam/gui/image_tool/azimuthal_integ_1d_view.py
JamesWrigley: Fixed in 31a6ca96.

extra_foam/pipeline/data_model.py
JamesWrigley: They weren't fully removed, they're originally declared in `DataItem` and this subclass re-declared them. That isn't recommended though so only the duplicate slots from the subclass are removed.

extra_foam/pipeline/processors/binning.py
JamesWrigley: Good catch, fixed in 31a6ca96.

CammilleCC (approved):
Thanks for implementing the nits :D LGTM

PR #310, opened 2021-07-06
Display crosshair in azimuthal integration window

### LINKED ISSUE #304: Display beam center in the azimuthal integration window
This would make it a little more user friendly.

JamesWrigley:
This also sets the default beam center to the center of the image. Feel free to suggest a better colorscheme/design, UIs are not my strong point :P

extra_foam_ai

Fixes #304.

### DIFF
--- a/extra_foam/gui/image_tool/azimuthal_integ_1d_view.py
+++ b/extra_foam/gui/image_tool/azimuthal_integ_1d_view.py
@@ -12,7 +12,7 @@ from PyQt5.QtWidgets import QVBoxLayout, QSplitter, QTabWidget
 from .base_view import _AbstractImageToolView, create_imagetool_view
 from ..ctrl_widgets import AzimuthalIntegCtrlWidget
 from ..misc_widgets import FColor
-from ..plot_widgets import ImageViewF, PlotWidgetF
+from ..plot_widgets import ImageViewF, PlotWidgetF, Crosshair
 from ...algorithms import find_peaks_1d
 from ...config import AnalysisType, plot_labels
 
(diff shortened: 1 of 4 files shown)


lucagelisio:
Looks great! :-)

lucagelisio (approved):
Perfect, approved from my side!

extra_foam/gui/image_tool/azimuthal_integ_1d_view.py
```diff
@@ -97,9 +100,20 @@ def initUI(self):
...
+
     def initConnections(self):
         """Override."""
-        pass
+        self._ctrl_widget.cx_changed_sgn.connect(lambda x: self.onBeamCenterChanged(x, None))
+        self._ctrl_widget.cy_changed_sgn.connect(lambda y: self.onBeamCenterChanged(None, y))
+
+    def onBeamCenterChanged(self, x, y):
+        if x is None:
+            x = float(self._ctrl_widget._cx_le.text())
```
CammilleCC, line 112: You could use `lineedit.value()` to avoid casting it to float. Seems that the smart edits casts to the appropriate type when a validator is added

CammilleCC (approved):
The crosshair looks nifty, good job!
We can pimp it up further by considering its transform corresponding to the view, such that its scale is invariant on zoom in/out. But that's not so important for now.

Comments aside, LGTM

extra_foam/gui/image_tool/azimuthal_integ_1d_view.py
JamesWrigley: Unfortunately that only works with some subclasses of `SmartLineEdit`, which have special `Validator` inner classes that `value()` uses to do the casting.

PR #311, opened 2021-07-22
Clean up dependencies and CMake configs, fix compatibility with Python 3.9

JamesWrigley:
This makes it easier to manage and upgrade the dependencies. My original motivation was to update xtensor to use a new feature, but it kinda metastasized into this :P 

Overview of changes (also see the commit messages):
- Move dependencies previously managed as submodules to Conda (762b45e0). This also simplifies installation.
- Clean up the CMake config; deleting unnecessary modules etc (762b45e0).
- Fix deprecation errors/warnings from Python 3.9 that I came across while testing (2527a643 and 09b9931f).
- Updated all of the C++ dependencies, there were some breaking changes in xtensor and TBB that needed fixing (a9b4df0d and 05dc0a7e). This also required upgrading to C++17.

To reviewers: I'd suggest reviewing each commit individually, they're fairly atomic and it'll make more sense than reviewing the whole thing at once. It would also be helpful if you try installing extra-foam from this branch following the new installation steps, configuring CMake is ludicrously complicated and it's possible I missed some edge cases from the previous configuration.


environment.yml
```diff
@@ -1,8 +1,19 @@
...
+  - cmake>=3.17
   - numpy>=1.19
-  - tbb-devel<2021.1
+  - tbb-devel>=2021
+  - tbb>=2021
+  - xtensor
+  - xtensor-python
+  - xtensor-blas
+  - xsimd
+  - pybind11
```
RobertRosca, line 19: Now that the GitHub Workflows are working it's worth creating a lock file with pinned dependencies and setting up Dependabot to automatically update them, this way we can automatically detect issues caused by dependency upgrades before they come up in some random PR or issue from a user.

This isn't really relevant to the PR, just wanted to write it now, I'll open a separate issue for this point.

RobertRosca:
LGTM :+1:

zhujun98:
This is for operation. Why do you want automatically dependency update?

RobertRosca:
The automatic dependency updates happen passively anyway since the constraints in `environment.yml` are mostly `>=`, so any time you do an installation it'll fetch the dependency greater than version X, usually giving you the latest version.

By having an explicit dependency lockfile that Dependabot can use to track updates you can trigger the CI whenever a dependency change happens and test it, you catch these issues before they randomly show up during a deployment or merge request, and also react to deprecations before they become an issue.

e.g. `numpy>=1.19` means any fresh install will get the latest numpy, if that breaks something then nobody with an existing environment would notice, you'd only find the issue during a deployment or during some random unrelated PR that triggers the CI and happens to pull in the new dependency.

It's super confusing to do a PR and see some random unrelated tests fail, or get an issue from a user describing a problem you can't reproduce on your local installation, until you realise that the problem isn't with a code change but is actually that the latest release of some random package is causing the issue in recent installations.

zhujun98:
> The automatic dependency updates happen passively anyway since the constraints in `environment.yml` are mostly `>=`, so any time you do an installation it'll fetch the dependency greater than version X, usually giving you the latest version.
> 
> By having an explicit dependency lockfile that Dependabot can use to track updates you can trigger the CI whenever a dependency change happens and test it, you catch these issues before they randomly show up during a deployment or merge request, and also react to deprecations before they become an issue.
> 
> e.g. `numpy>=1.19` means any fresh install will get the latest numpy, if that breaks something then nobody with an existing environment would notice, you'd only find the issue during a deployment or during some random unrelated PR that triggers the CI and happens to pull in the new dependency.

`numpy` is a low risk dependency. But others, like `xtensor` series, TBB, even `Pillow` are high risk dependencies. In practice, you won't need the new features most of the time. It is why only fixed versions should be used for most dependencies. Automatically pulling new versions should not be allowed for most dependencies.

In a word, the solution is good but the problem itself can be avoided.

> It's super confusing to do a PR and see some random unrelated tests fail, or get an issue from a user describing a problem you can't reproduce on your local installation, until you realise that the problem isn't with a code change but is actually that the latest release of some random package is causing the issue in recent installations.

I have never experienced this issue before. Again, only automatically update low-risk dependency can save the trouble.

JamesWrigley:
> `numpy` is a low risk dependency. But others, like `xtensor` series, TBB, even `Pillow` are high risk dependencies. In practice, you won't need the new features most of the time. It is why only fixed versions should be used for most dependencies. Automatically pulling new versions should not be allowed for most dependencies.

I think that for most of these libraries breaking changes happen fairly rarely. This time the only breakages came from versions of TBB and xtensor that are both ~1.5 years old, that rate of change is ok with me. In general I am in favour of keeping all our dependencies up-to-date because sooner or later we will have to upgrade them anyway, and I would prefer to discover all the breakage as soon as possible rather than all at once. This is partly personal preference though. @CammilleCC, what do you think?

> I have never experienced this issue before. Again, only automatically update low-risk dependency can save the trouble.

This has happened to me many times, and it is pretty confusing :confused:

zhujun98:
You guys make the decision.

> This has happened to me many times, and it is pretty confusing confused

I just wonder how could this happen if your development / test environment is the same as the deploy environment. But never mind ...

JamesWrigley:
In my experience, development environments are rarely exactly the same as deployment/test environments. I almost never update my local conda environment for example (out of laziness rather than intent). So if someone's made a PR and the CI pulls in a new version of a dependency that breaks something, then that test will fail for (probably) completely unrelated reasons to the PR.

Of course that would be not be a problem if we pin all the dependency versions, but that's not my personal preference for the reasons above.

zhujun98:
> So if someone's made a PR and the CI pulls in a new version of a dependency that breaks something, then that test will fail for (probably) completely unrelated reasons to the PR.

It is why high-risk dependency should not be automatically updated/pulled!

RobertRosca:
I think I didn't explain this very well. Right now out of the dependencies in `setup.py` and `environment.yml` the only pinned dependencies are:

```
'Pillow==8.2.0',
'PyQt5==5.13.2',
'hiredis==1.0.1',
'imageio==2.8.0',
'redis==3.5.2',
'tbb-devel<2021.1',
```

So any time you install extra-foam you'll get the latest (compatible) version of all of the other dependencies, as well as all of their dependencies. Took a quick look and that ends up being:

```
- cfelpyutils
- cycler
- EXtra-data
- EXtra-geom
- fabio
- h5py
- karabo-bridge
- kiwisolver
- matplotlib
- msgpack
- msgpack-numpy
- numexpr
- numpy
- pandas
- psutil
- pyFAI
- pyparsing
- pytz
- PyYAML
- pyzmq
- scipy
- silx
- six
- toolz
- xarray
```

Every time a new installation is done all of these may change, since they aren't explicitly locked by us.

> > This has happened to me many times, and it is pretty confusing confused
>
> I just wonder how could this happen if your development / test environment is the same as the deploy environment. But never mind ...

If you set up your development environment and the next day one of the packages listed above has an update which causes an issue/change in behaviour then you'd only notice when a fresh installation is done by the CI, a new deployment, or somebody else.

>  is why high-risk dependency should not be automatically updated/pulled!

But they are. Right now the newest versions are pulled in for the majority of the dependencies without any indication or record of what was installed before, since only 6 out of ~30 dependencies are pinned (this isn't entirely true since some of the indirect dependencies may be constrained by the direct dependencies, but that isn't managed by us, so we don't control those constraint changes).

It's not that I'm suggesting we add in a change to have automatically updating dependencies, that's the current behaviour for the majority of them already, I'm suggesting adding in a lockfile so that the CI is trigger whenever one of these updates happens so that any problems can be automatically caught instead of only noticed when a fresh installation is done.

CammilleCC (approved):
Wow, big changes! Thanks for the all the work :) I appreciate how the commits are well-organized, changes are easier to track.

While I am happy to have dynamic dependency versions to make `extra-foam` more "up-to-date", I understand Jun's concerns. It would be best to follow up this PR with a dependency check, as already mentioned by Robert. Either with a `lockfile` as previously mentioned, or a nightly build. I am inclined with the latter as it is more periodic :P 

LGTM

zhujun98:
> It's not that I'm suggesting we add in a change to have automatically updating dependencies, that's the current behaviour for the majority of them already, I'm suggesting adding in a lockfile so that the CI is trigger whenever one of these updates happens so that any problems can be automatically caught instead of only noticed when a fresh installation is done.

Thanks for the explanation @RobertRosca! As I said at the very beginning, the solution is very good.

Also thanks for the impressive work done by @JamesWrigley.

PR #315, opened 2021-09-08
Save the pump-probe output to file

CammilleCC:
SPB requested to have a saved file for the pump-probe values. We now offer an ASCII and an NPZ output.

### DIFF
--- a/extra_foam/gui/ctrl_widgets/pump_probe_ctrl_widget.py
+++ b/extra_foam/gui/ctrl_widgets/pump_probe_ctrl_widget.py
@@ -65,6 +65,8 @@ class PumpProbeCtrlWidget(_AbstractCtrlWidget):
         self._abs_difference_cb.setChecked(True)
 
         self._reset_btn = QPushButton("Reset")
+        self.save_btn = QPushButton("Save to file")
+        self.save_btn.setFixedWidth(150)
 
         self.initUI()
         self.initConnections()
@@ -88,6 +90,7 @@ class PumpProbeCtrlWidget(_AbstractCtrlWidget):
         layout.addWidget(self._off_pulse_le, 1, 3)
 
         layout.addWidget(self._abs_difference_cb, 0, 4, AR)
+        layout.addWidget(self.save_btn, 2, 4, AR)
 
         self.setLayout(layout)
 
--- a/extra_foam/gui/windows/pump_probe_w.py
+++ b/extra_foam/gui/windows/pump_probe_w.py
@@ -7,6 +7,7 @@ Author: Jun Zhu <jun.zhu@xfel.eu>
 Copyright (C) European X-Ray Free-Electron Laser Facility GmbH.
 All rights reserved.
 """
+from enum import Enum
 import os
 import os.path as osp
 
@@ -95,6 +96,11 @@ class PumpProbeFomPlot(TimedPlotWidgetF):
         self._plot.setData(x, y)
 
 
+class SaveFile(Enum):
+    NPZ = "NumPy Binary File (*.npz)"
+    TXT = "Text File (*.txt)"
+
+
 class PumpProbeWindow(_AbstractPlotWindow):
     """PumpProbeWindow class."""
     _title = "Pump-probe"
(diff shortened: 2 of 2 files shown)


dallanto:
I could have sworn that this MR is open for a longer time, at least I'm sure I've looked at the code soon after you committed in July. LGTM 🙂

extra_foam/gui/ctrl_widgets/pump_probe_ctrl_widget.py
```diff
@@ -65,6 +65,8 @@ def __init__(self, *args, **kwargs):
         self._abs_difference_cb.setChecked(True)
 
         self._reset_btn = QPushButton("Reset")
+        self.save_btn = QPushButton("Save to file")
```
ebadkamil, line 68: would rather prefer a QAction in a toolbar on this window instead of a button to save to files.

extra_foam/gui/windows/pump_probe_w.py
```diff
@@ -138,8 +148,61 @@ def initUI(self):
...
+        self._ctrl_widget.save_btn.clicked.connect(self.saveToFile)
 
     def closeEvent(self, QCloseEvent):
         self._ctrl_widget.resetAnalysisType()
         super().closeEvent(QCloseEvent)
+
+    def saveToFile(self):
+        # Get the current data
+        if not len(self._queue):
+            return
```
JamesWrigley, line 160: It would be better if the button itself was disabled when there is no data to save, rather than silently exiting.

extra_foam/gui/windows/pump_probe_w.py
```diff
@@ -138,8 +148,61 @@ def initUI(self):
...
+        data = self._queue[0]
```
JamesWrigley, line 161: This changed in #318, the GUI now receives a `dict` of all the data from the workers. So here we'd now need `self._queue[0]["processed"]`.

extra_foam/gui/windows/pump_probe_w.py
```diff
@@ -138,8 +148,61 @@ def initUI(self):
...
+
+        # Validate filepath
+        if not filepath:
+            return
+        suffix = f".{filter.name.lower()}"
+        if not filepath.lower().endswith(suffix):
+            filepath += suffix
+
+        # Copy reference file from tmp folder to desired destination
+        os.makedirs(osp.dirname(filepath), exist_ok=True)
```
JamesWrigley, line 178: This comment doesn't seem to match the code? And why is it necessary to use `os.makedirs()` here? I thought we could trust that any directory returned by `QFileDialog.getSaveFileName()` already exists since the user would need to create it first in the dialog.

extra_foam/gui/windows/pump_probe_w.py
```diff
@@ -138,8 +148,61 @@ def initUI(self):
...
+        # Save the data
+        pp = data.pp
+        train_id = data.tid
+        position = pp.x
+        intensity_on = pp.y_on
+        intensity_off = pp.y_off
+        intensity_sub = pp.y
+
+        if position is None:
+            return
```
JamesWrigley, line 189: I think this should show an error dialog to make it clear to the user that nothing happened.

extra_foam/gui/windows/pump_probe_w.py
```diff
@@ -138,8 +148,61 @@ def initUI(self):
...
+        if filter is SaveFile.NPZ:
+            np.savez(filepath,
+                     trainId=train_id,
+                     position=position,
+                     intensity_on=intensity_on,
+                     intensity_off=intensity_off,
+                     intensity_subtracted=intensity_sub)
+        elif filter is SaveFile.TXT:
+            # write out conversion to tabular ASCII
+            with open(filepath, 'w') as f:
```
JamesWrigley, line 200: Could we use Python's `csv` library here?

JamesWrigley (changes requested):

JamesWrigley:
Apologies for my shockingly slow review :see_no_evil:

extra_foam/gui/ctrl_widgets/pump_probe_ctrl_widget.py
JamesWrigley: Hmm, @CammilleCC I leave this up to you. I think either UI is acceptable.

PR #316, opened 2021-09-16
Update pyzmq to fix a memory leak

### LINKED ISSUE #313: Memory leak with certain detectors
When streaming runs from a file, the memory usage of the pulse worker will steadily increase until it gets killed by the OOM killer. So far it's only been observed with AGIPD data, DSSC and EPIX work fine.

One fix is to explicitly call the garbage collector in the main loop of [`ProcessWorker.run()`](https://github.com/European-XFEL/EXtra-foam/blob/dev/extra_foam/pipeline/f_worker.py#L107) (similar to ebe83608), but this clobbers performance. I'd guess that the pulse worker is doing something that creates a bunch of objects that are never cleaned, perhaps during assembly or something? It's strange that this wasn't noticed before, since extra-foam instances are left running for days at a time during experiments.

JamesWrigley:
See the commit message for details. This is the script I used to reproduce the issue outside of extra-foam:
```python
#! python3

import gc
import sys
import signal

import zmq
from karabo_bridge import Client


def exit_handler(*_):
    print("\nExiting")
    sys.exit(0)

if __name__ == "__main__":
    signal.signal(signal.SIGINT, exit_handler)

    client = Client("tcp://127.0.0.1:45454")

    while True:
        client.next()
        # gc.collect()
        print("Received data")
```

Fixes #313.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -299,7 +299,7 @@ setup(
         'scipy>=1.2.1',
         'msgpack>=0.5.6',
         'msgpack-numpy>=0.4.4',
-        'pyzmq>=17.1.2',
+        'pyzmq>=22.3.0',
         'pyFAI>=0.17.0',
         'PyQt5==5.13.2',
         'PyQt5-sip>=12.9.0',


JamesWrigley:
One thing I forgot to mention in the commit message was where we use the `buffer` attribute mentioned in the pyzmq bug report: https://github.com/European-XFEL/karabo-bridge-py/blob/master/karabo_bridge/serializer.py#L199

CammilleCC (approved):
LGTM, thanks for looking into this :)

tmichela (approved):

PR #318, opened 2021-10-04
Add support for streaming detector data and normalizing by user-defined 2D pipeline sources

JamesWrigley:
There was a request from MID to add support for normalizing stuff by the ePix detector, I decided that the best way to implement this in extra-foam would be by streaming the processed detector data from an ePix instance over a Karabo bridge to the AGIPD instance. This required a number of changes, and I've tried to organize them by commits to make review easier. MID used this successfully during their last experiment.

## How to run it for testing
This is not so easy :grimacing: What you really want is to run a `PipeToZeroMQ` device with two outputs, but in lieu of this I wrote a (truly disgusting) [script](https://gist.github.com/JamesWrigley/62203edd06f12bf66af4cad36cf1954d) that would stream the data from a run to separate ports. In addition, I added a section to the extra-foam config to add a new bridge for the ePix extra-foam instance:
```yml
# Just plonk this somewhere in <instrument>.config.yaml
STREAMER:
    ZMQ:
        EF_ePix100:
          DEFAULT_TYPE: 0
          ADDR: 127.0.0.1
          PORT: 5558
```

With this done, the steps to get the whole thing running are:
1. Start the bridge server:
    ```bash
    $ karabo-bridge-serve-ghidorah
    ```
2. Open the extra-foam instances for the ePix and AGIPD:
    ```bash
    $ extra-foam agipd mid
    $ extra-foam epix mid # Run this in a different terminal
    ```
3. Start the ePix instance.
4. Add a source for the ePix to the AGIPD instance in the 'User-defined' section (see docs)
5. Check that the port of the ePix extension and the source port in the AGIPD instance match, then start the AGIPD instance.
6. The ePix should be listed as an available source under the ROI normalization settings.

You could do this all on Maxwell, but you could also do it locally with a truncated run. I've been testing with proposal 2929, run 149 and I have a truncated 7.7GB version of it lying around that I use for testing (feel free to ping me if you'd like it).

Oversized screenshot of the whole shebang (AGIPD instance on the left, ePix on the right):
image

### DIFF
--- a/extra_foam/config.py
+++ b/extra_foam/config.py
@@ -32,6 +32,11 @@ class DataSource(IntEnum):
     UNKNOWN = 2  # not specified
 
 
+class KaraboType(IntEnum):
+    CONTROL_DATA = 0
+    PIPELINE_DATA = 1
+
+
 class PumpProbeMode(IntEnum):
     UNDEFINED = 0
     REFERENCE_AS_OFF = 1  # use pre-defined reference image
--- a/extra_foam/config.py
+++ b/extra_foam/config.py
@@ -7,7 +7,7 @@ Author: Jun Zhu <jun.zhu@xfel.eu>
 Copyright (C) European X-Ray Free-Electron Laser Facility GmbH.
 All rights reserved.
 """
-from enum import IntEnum
+from enum import Enum, IntEnum
 import os.path as osp
 import shutil
 from collections import abc, namedtuple, OrderedDict
@@ -24,6 +24,11 @@ from .logger import logger
 
 _MAX_INT32 = np.iinfo(np.int32).max
 _MIN_INT32 = np.iinfo(np.int32).min
+
+
+class ExtensionType(Enum):
+    ALL_OUTPUT = "endpoint"
+    DETECTOR_OUTPUT = "detector_endpoint"
 
 
 class DataSource(IntEnum):
--- a/extra_foam/gui/ctrl_widgets/data_source_widget.py
+++ b/extra_foam/gui/ctrl_widgets/data_source_widget.py
@@ -26,7 +26,7 @@ from ..gui_helpers import parse_boundary, parse_slice
 from ..misc_widgets import FColor
 from ..mediator import Mediator
 from ...database import MonProxy
-from ...config import config, DataSource
+from ...config import config, DataSource, KaraboType
 from ...geometries import module_indices
 from ...processes import list_foam_processes
 from ...logger import logger
@@ -395,7 +395,7 @@ class DataSourceItemModel(QAbstractItemModel):
                 for ppt in ppts:
                     # For now, all pipeline data are exclusive
                     src_item = DataSourceTreeItem(
-                        [False, 1, src, ppt, default_slicer, default_v_range],
+                        [False, KaraboType.PIPELINE_DATA, src, ppt, default_slicer, default_v_range],
                         exclusive=True,
                         parent=ctg_item)
                     ctg_item.appendChild(src_item)
@@ -416,7 +416,7 @@ class DataSourceItemModel(QAbstractItemModel):
             for src, ppts in srcs.items():
                 for ppt in ppts:
                     ctg_item.appendChild(DataSourceTreeItem(
-                        [False, 0, src, ppt, default_slicer, default_v_range],
+                        [False, KaraboType.CONTROL_DATA, src, ppt, default_slicer, default_v_range],
                         exclusive=False,
                         parent=ctg_item))
 
@@ -701,10 +701,9 @@ class DataSourceWidget(_AbstractCtrlWidget):
             if isinstance(v, int):
                 painter.setPen(Qt.NoPen)
 
-                if v == 0:
-                    # control data
+                if v == KaraboType.CONTROL_DATA:
                     painter.setBrush(self._c_brush)
-                elif v == 1:
+                elif v == KaraboType.PIPELINE_DATA:
                     # pipeline data
                     painter.setBrush(self._p_brush)
                 else:
--- a/extra_foam/gui/ctrl_widgets/data_source_widget.py
+++ b/extra_foam/gui/ctrl_widgets/data_source_widget.py
@@ -439,8 +439,9 @@ class DataSourceItemModel(QAbstractItemModel):
                                       parent=self._root)
         self._root.appendChild(ctg_item)
         for i in range(n_user_defined):
+            dtype = KaraboType.PIPELINE_DATA if i < n_user_defined / 2 else KaraboType.CONTROL_DATA
             ctg_item.appendChild(DataSourceTreeItem(
-                [False, 0, f"Device-ID-{i+1}", f"Property-{i+1}",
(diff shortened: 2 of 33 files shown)


extra_foam/pipeline/f_worker.py
```diff
@@ -152,6 +156,26 @@ def run(self):
...
+                    processed = data_out["processed"]
+                    detector, key = data_out["catalog"].main_detector.split()
+
+                    now = time.time()
+                    bridge_data = {
+                        f"EF_{detector}" : {
+                            key: processed.image.masked_mean,
+                            "metadata": {
+                                "timestamp": now,
+                                "timestamp.sec": int(now),
```
JamesWrigley, line 169: I don't know if this is a good idea. The idea was to make it match the time we're finished with processing in extra-foam, but maybe that's dumb since it won't match the timestamp from the train ID.

extra_foam/pipeline/f_transformer.py
```diff
@@ -70,8 +70,7 @@ def transform_euxfel(data, *, catalog=None, source_type=DataSource.UNKNOWN):
 
             if modules:
                 prefix, suffix = src_name.split("*")
-                new_raw[src] = dict()
```
CammilleCC, line 73: Would it be alright if we don't have a default value for `src` if `n_found == 0`?

extra_foam/config.py
```diff
@@ -26,12 +26,22 @@
 _MIN_INT32 = np.iinfo(np.int32).min
 
 
+class ExtensionType(Enum):
```
CammilleCC, line 29: I appreciate using string values here :D

extra_foam/config.py
```diff
@@ -26,12 +26,22 @@
...
+    DETECTOR_OUTPUT = "detector_endpoint"
+
+
 class DataSource(IntEnum):
     FILE = 0  # data from files (run directory)
     BRIDGE = 1  # real-time data from the bridge
     UNKNOWN = 2  # not specified
 
 
+class KaraboType(IntEnum):
```
CammilleCC, line 40: ...and also creating a new enum. Thanks!

extra_foam/gui/ctrl_widgets/data_source_widget.py
```diff
@@ -439,8 +439,9 @@ def setupModelData(self):
                                       parent=self._root)
         self._root.appendChild(ctg_item)
         for i in range(n_user_defined):
+            dtype = KaraboType.PIPELINE_DATA if i < n_user_defined / 2 else KaraboType.CONTROL_DATA
```
CammilleCC, line 442: Could you expound the condition here please?

extra_foam/gui/ctrl_widgets/extension_ctrl_widget.py
```diff
@@ -62,7 +72,22 @@ def loadMetaData(self):
         """Override."""
         pass
 
+    def _getAvailablePort(self, default_port):
```
CammilleCC: Nice addition! Would be great if this is in a module that can be also accessed by the special suite codes

extra_foam/pipeline/f_worker.py
CammilleCC: I believe it would be alright. IIRC matching is based on `tid` and not on the timestamp. If it is possible, you could add a matching test on `test_transformer` for a deferred update with a same `tid` and a different timestamp.

CammilleCC (approved):
Looks great in general! I did a quick look on the code and please see below some of my comments. 

I haven't had a good chance to play around with some actual streams, but the added tests looks sufficient.

extra_foam/pipeline/f_transformer.py
JamesWrigley: Unless I'm misunderstanding, I think that's exactly what my change implements? Later on there's a check for `n_found == 0`, and only if that is not the case is the dict stored in `new_raw`.

extra_foam/gui/ctrl_widgets/data_source_widget.py
JamesWrigley: This is checking if the new source is in the first or last half, if it's the first half then it's type is set to pipeline data, and if it's the last half then it'll be control data. ATM extra-foam is configured to create four user-defined sources so two will be for pipeline data and two for control data.

extra_foam/gui/ctrl_widgets/extension_ctrl_widget.py
JamesWrigley: Sure, but why? I don't think any special suite creates a server :thinking:

extra_foam/pipeline/f_worker.py
JamesWrigley: On second thoughts, maybe the timestamp should just not be set at all? I feel like it's kinda confusing to have the timestamp not match the train ID, since I think everywhere else at XFEL it is assumed that those match.

extra_foam/gui/ctrl_widgets/extension_ctrl_widget.py
CammilleCC: I concur, it's just that I don't feel that this method belongs in this class, even though it is only used here. Perhaps in `extra.foam.utils` or something similar?

extra_foam/pipeline/f_transformer.py
CammilleCC: I see, I am who misunderstood. Thanks for the clarification :D

extra_foam/gui/ctrl_widgets/extension_ctrl_widget.py
JamesWrigley: Ah, sure. Fixed in c9dc112d.

extra_foam/pipeline/f_worker.py
JamesWrigley: Decided to remove this in bc79978e, I think it's just too weird to have the other fields of the timestamp not match the train ID.

tmichela: I don't know the exact purpose here, but it is fine that the timestamp does not match the trainId (in karabo at least). in karabo this is metadata information about the pipeline data. by default it takes the time when it is written to an output channel: if you have a data going through multiple devices, then the tid/timestamp are overwritten everytime here.

You should rely on the timestamp information in the data if that is important.

For example in the train matcher, for pipeline data we replicate the source metadata (trainId and timestamp), but for control data, since they are originally distributed via message broker, we don't have this metadata information, so we create a metadata for it when  it is sent out. In this new metadata, the trainId is taken from the source but the timestamp is the one when it is sent out. But the real timestamp (when the property value was updated is contained in the data)

JamesWrigley: Ah hah, interesting :thinking: EXtra-foam doesn't rely on the timestamp, I only put it in there because it's part of the karabo-bridge protocol. But maybe it's better to match what Karabo does, if only for the sake of consistency.

JamesWrigley: Reverted in 0d816ce5.

PR #319, opened 2021-10-05
Clean up dependencies and configure Dependabot

### LINKED ISSUE #312: Conda Lockfile and Dependabot Integration
Now that the GitHub Workflows are working it's worth creating a lock file with pinned dependencies and setting up Dependabot to automatically update them, this way we can automatically detect issues caused by dependency upgrades before they come up in some random PR or issue from a user.

_Originally posted by @RobertRosca in https://github.com/European-XFEL/EXtra-foam/pull/311#r675137066_

JamesWrigley:
Partially fixes #312. Dependabot doesn't support Anaconda though so I moved all the pip-available deps to `setup.py`, and for now I've pinned the other dependencies that might break.


.github/workflows/tests.yml
```diff
@@ -12,8 +12,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: [3.7]  #  TODO: Enable additional versions: 3.7
-        gcc-version: [8]  #  TODO: Enable additional versions: 6, 7
```
CammilleCC, line 16: Could you please remind me why gcc-related installations are removed? Are these not being used anymore in favor of gxx?

CammilleCC (approved):
I am not the best person to consult with dependencies-related changes, but these seem quite alright. TIL about `pip constraints`.

L(!)GTM

.github/workflows/tests.yml
JamesWrigley: Yeah, now the compilers are installed with conda so this isn't necessary. I actually changed that a while ago, deleting these lines is just cleanup.

setup.py
```diff
@@ -295,25 +295,34 @@ def has_ext_modules(self):
...
+        # These dependencies are not directly used, but are needed to satisfy
+        # pip's resolver:
+        'pygments',
+        'jinja2',
+        'decorator',
+        'pexpect>4.3',
+        'pygments',
+        'toolz>=0.8.2',
+        'packaging',
+        'cffi>=1'
```
RobertRosca: You don't actually need to do this for dependabot to work, since the configuration is:

https://github.com/European-XFEL/EXtra-foam/blob/a93a9540ca609e164a532a26afa8fd2391a6a38b/.github/dependabot.yml#L1-L6

Where the `.github/dependabot/` directory only has the `constraints.txt` file in it.

And the install step for the tests is:

https://github.com/European-XFEL/EXtra-foam/blob/a93a9540ca609e164a532a26afa8fd2391a6a38b/.github/workflows/tests.yml#L62

That means that dependabot will bump the constraints files, create a PR, and the tests execute with the bumped dependency versions.

JamesWrigley: Ah hah, good point. Fixed in da95b2e7.

RobertRosca:
LGTM :+1:

PR #321, opened 2021-10-29
Add support for special suites supporting both EXtra-foam and Karabo

JamesWrigley:
Some special suites may want to support processing data from both EXtra-foam and
Karabo bridges, this is now possible by passing ClientType.BOTH to the
constructor of _SpecialAnalysisBase(). More generally, this replaces the
`client_klass` parameter of create_special() with an argument to the
_SpecialAnalysisBase() constructor.

If both client types are supported, then a combo box will be available for the
user to select the appropriate source.

### DIFF
--- a/extra_foam/special_suite/cam_view_w.py
+++ b/extra_foam/special_suite/cam_view_w.py
@@ -22,7 +22,7 @@ from .cam_view_proc import (
     CamViewProcessor, _DEFAULT_N_BINS, _DEFAULT_BIN_RANGE
 )
 from .special_analysis_base import (
-    create_special, QThreadKbClient, _BaseAnalysisCtrlWidgetS,
+    create_special, ClientType, _BaseAnalysisCtrlWidgetS,
     _SpecialAnalysisBase
 )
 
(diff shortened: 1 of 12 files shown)


CammilleCC (approved):
This is a cool addition! I would suggest putting the `client_support` as class variable, or would you foresee changes on the instantiation itself?

Aside from that, LGTM

JamesWrigley:
No, I just want to make `create_special()` unnecessary :P IMHO it's overcomplicated and I want to get rid of it at some point, so that's why I made `client_support` a regular parameter. Is there an advantage to making it a class variable?

CammilleCC:
I see, then I support you with that endeavor :P

It's more of flavor, TBH. I oought to use class variables as annotation (similar with title and long title). Using it also somewhat makes code cleaner because it reduces passed arguments and variable creation on init. But it's of course up to you :D

JamesWrigley:
Actually that's a good point about `_[long_]title`. I was hesitant to make it a class variable because it means that inheriting classes would need to know about that field (that's the advantage of `create_special()`) but I forgot that the title fields already do this. Made `_client_support` a class field in 315008fb.

PR #322, opened 2021-10-29
Assorted minor GUI fixes

JamesWrigley:
This is just a collection of minor changes I've made over the last couple months, mostly related to the GUI (except for ca6cbe18).

### DIFF
--- a/extra_foam/gui/gui_helpers.py
+++ b/extra_foam/gui/gui_helpers.py
@@ -10,7 +10,7 @@ All rights reserved.
 import os.path as osp
 
 from PyQt5.QtCore import QSize
-from PyQt5.QtGui import QIcon
+from PyQt5.QtGui import QIcon, QGuiApplication, QCursor
 from PyQt5.QtWidgets import QPushButton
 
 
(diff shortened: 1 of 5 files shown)


extra_foam/gui/gui_helpers.py
```diff
@@ -230,3 +230,22 @@ def invert_dict(mapping):
...
+    :param bool    resize: Whether or not to resize the window.
+    """
+    screen = QGuiApplication.screenAt(QCursor.pos())
+
+    if resize:
+        screen_size = screen.size()
+        max_width = screen_size.width()
+        max_height = screen_size.height()
+
+        window.resize(max_width * 0.8, max_height * 0.8)
```
CammilleCC, line 249: This is a good addition :) while it is not used right now, we may have to check the windows' behavior on 4k and 720p monitors in the future.

extra_foam/gui/plot_widgets/image_view_base.py
```diff
@@ -120,14 +123,20 @@ def __init__(self, *,
...
+            layout.addWidget(self._hist_widget, 1)
+
         layout.setSpacing(0)
         layout.setContentsMargins(0, 0, 0, 0)
         self.setLayout(layout)
 
     def reset(self):
         self.clear()
 
+    def addMenuItem(self, *args, **kwargs):
```
CammilleCC: I believe it would be better to follow the signature of the parent `addMenuItem`, rather than using `*args, **kwargs`

CammilleCC (commented):
Aside from the minor comments, LGTM

extra_foam/gui/plot_widgets/image_view_base.py
JamesWrigley: Fixed in 30eb8d96.

CammilleCC (approved):
Thanks for considering the suggestion! LGTM

extra_foam/special_suite/facade.py
```diff
@@ -70,6 +70,9 @@ def __init__(self, topic):
         self._cw = QWidget()
         self.setCentralWidget(self._cw)
 
+        # Center window
```
ebadkamil: Remove the comment. Function name is descriptive enough.

JamesWrigley: Fixed in 0d6165db.

PR #323, opened 2021-11-25
Convert floats to ints for Python 3.10 compatibility

JamesWrigley:
Python 3.10 included this change: https://bugs.python.org/issue37999. Turns out
that extra-foam's copy of pyqtgraph relies on this in some places, which results
in `TypeError`'s when running it with Python 3.10. These are just the places in
the code that I could find with manual/unit tests, it's possible that there are
more. The plan is to replace the internal copy with upstream though, so I don't
think it's worth doing anything more rigorous.

Note: maybe we should pin the Python version?

### DIFF
--- a/extra_foam/gui/pyqtgraph/Point.py
+++ b/extra_foam/gui/pyqtgraph/Point.py
@@ -158,4 +158,4 @@ class Point(QtCore.QPointF):
         return Point(self)
         
     def toQPoint(self):
-        return QtCore.QPoint(*self)
+        return QtCore.QPoint(int(self.x()), int(self.y()))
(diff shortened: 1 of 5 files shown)


JamesWrigley:
Found another one in 5afd5b64, also fixed upstream in https://github.com/pyqtgraph/pyqtgraph/pull/1716. We really gotta switch to upstream...

CammilleCC (approved):
I also upgraded Python and checked for a bit, and I believe that this changes would suffice for now. LGTM.

I don't mind not pinning Python version, but we should consider adding more versions in the CI.

JamesWrigley:
I thought about that, but I don't see much point in adding more versions in the CI. We already control the Python version with Conda, so I would rather just keep compatibility with a single version at a time. And I think the CI is already using Python 3.10 because we haven't pinned a version.

Unfortunately, in this case the CI wouldn't have caught it anyway, because we don't run the pyqtgraph tests :upside_down_face: (assuming their tests cover these code paths)

PR #328, opened 2021-12-10
Implement the Correlator special suite

JamesWrigley:
Finally got around to making a PR out of this :sweat_smile: I'm keeping it as a draft for now because there are a couple of (hopefully) minor issues still to fix:
- f581c873 needs to be refactored, you can ignore this commit for now.
- This uses the [`high_high_water_mark`](https://git.xfel.eu/karaboDevices/metropc/-/tree/high_high_water_mark) branch of `metropc`, which in turn is based on [`feat/bound_device_features`](https://git.xfel.eu/karaboDevices/metropc/-/tree/feat/bound_device_features). The `high_high_water_mark` branch just increases the HWM for the output socket of `metropc`, because I noticed that sometimes `metropc` was just dropping outputs and that fixed the issue. It could also be because the special suite is not reading fast enough, but that needs to be investigated. And perhaps we should wait to merge this until the `feat/bound_device_features` branch is merged too.

To reviewers, even though this is a draft I would suggest starting now because there's... a lot... of code. There's also some auxiliary commits that were needed for the Correlator, so I would suggest reviewing each commit individually because they're fairly atomic (it'll make more sense that way).

~~Note: this PR depends on #326, so you should ignore commits 1581c1b2 and 479e93a8.~~


JamesWrigley:
Hmm, it's failing because the CI can't install `metropc`. Currently it's using an SSH URL but the repo is private anyway so using HTTPS won't help. @philsmt, is there a timeline to make `metropc` public?

extra_foam/utils.py
```diff
@@ -277,3 +281,96 @@ def get_available_port(default_port):
...
+                               (in which case the series label will be the
+                               keyword argument name), or a :code:`Series` object. For
+                               example, :code:`y1=42` will create a series with the
+                               name 'y1', and :code:`y42=Series(1, name='foo')` will
+                               create a series named 'foo' (optional).
+    :return:                   A :code:`DataArray` containing the data and
+                               metadata for plotting.
+    """
+    xr_attrs = {
+        "xlabel": xlabel,
```
philsmt, line 356: When receiving data as an `xarray.DataArray` (as you create here), there is already a documented [protocol](https://github.com/philsmt/metro-sci/blob/master/src/metro/devices/display/fast_plot.py#L1222) in `metropc` to use the coordinate label(s) as X axis label in 1D, whereas the y label is passed as an attribute (like here).

Would you mind using this instead of an `xlabel` attribute for compatibility?

JamesWrigley: Yeah sure, that should be fine.

philsmt:
We should probably aim to point this at `metropc`'s `main` branch 🤔

JamesWrigley:
I think you pressed the wrong button there :smile:

philsmt:
I did, apologies 😳

PR #387, opened 2022-01-31
Fix setting the center-of-mass

JamesWrigley:
It looks like a rebase went wrong here and put some functions in the wrong
spot, which caused the center-of-mass to not be set by updateF().


JamesWrigley:

PR #400, opened 2022-02-02
Update the docs

JamesWrigley:
Major changes:
- Remove mentions of the `PipeToZeroMQ` device.
- Rewrite the data flow tutorial to be about streaming data from Karabo.
- Properly link PRs in the changelog.
- Add a troubleshooting section for the trainmatcher and EXtra-foam.

Suggested way to see the docs if you don't want to bother installing the whole of EXtra-foam:
```bash
$ pip install sphinx nbsphinx ipython sphinx_rtd_theme sphinxcontrib_github_alt
$ cd docs
$ make html
```

I'd also recommend reviewing each commit one-by-one, they're quite atomic so the changes might make more sense that way.


JamesWrigley:
Any objections to moving the trainmatcher/EXtra-foam troubleshooting stuff to the COPS wiki? I didn't think about that while writing this, and it's probably better to keep everything in one place as much as possible. We could just link to the COPS from EXtra-foams docs.

tmichela:
I think COPS are only readable for DATA people, if you want other people to read it, we should maybe do the other way around (documentation here, links in redmine)?

I guess it's also more likely to have the doc attached to the project being up to date than on redmine 🤷

JamesWrigley:
True true, the other way around works too. And that way we get the nice formatting of RTD :stuck_out_tongue:

JamesWrigley:

PR #428, opened 2022-02-16
Require a recent compiler

JamesWrigley:
Otherwise we get compilation errors with jemalloc.


JamesWrigley:

PR #436, opened 2022-02-24
Add support for streaming runs by proposal and run number

JamesWrigley:
This is very convenient for runs on Maxwell.
Note: might have to revisit the GUI for this since it's getting a bit cluttered.

image


JamesWrigley:
If there's no objections, I'll merge this in a few days.

PR #563, opened 2022-07-20
Gracefully handle 0s when plotting in log scale

JamesWrigley:
Previously the behaviour was to add 1 before taking the log to avoid 0s causing
inf's in the output, but in some cases with very small values that gives wildly
incorrect results. Now we explicitly only take the log of non-zero values.

Before:
image

After:
image

### DIFF
--- a/extra_foam/gui/pyqtgraph/graphicsItems/GraphicsObject.py
+++ b/extra_foam/gui/pyqtgraph/graphicsItems/GraphicsObject.py
@@ -129,7 +129,7 @@ class PlotItem(GraphicsObject):
         """Convert array result to logarithmic scale."""
         ret = np.nan_to_num(arr)
         ret[ret < 0] = 0
-        return np.log10(ret + 1)
+        return np.log10(ret, where=(ret > 0))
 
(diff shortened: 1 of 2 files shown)


extra_foam/gui/pyqtgraph/graphicsItems/GraphicsObject.py
```diff
@@ -129,7 +129,7 @@ def toLogScale(arr, policy=None):
         """Convert array result to logarithmic scale."""
         ret = np.nan_to_num(arr)
         ret[ret < 0] = 0
-        return np.log10(ret + 1)
+        return np.log10(ret, where=(ret > 0))
```
zhujun98, line 132: I agree that the original solution does not work in some situations, as you have shown. However, there are also corner cases for the new one. It makes 1 and 0 the same. 0 will not be affected while 1 will become 0 after the log operation. Imaging you have a step function `f(x) = 1 if x > 0 else 0`, with the new solution the curve will become completely flat (all 0). Therefore, a better solution might be letting users choose among different methods.

PR #605, opened 2023-09-18
Add support for channel 2 of AdqDigitizer's

JamesWrigley:
I also sneaked in 6d911682 to fix the tests :hand_over_mouth:

### DIFF
--- a/extra_foam/pipeline/processors/digitizer.py
+++ b/extra_foam/pipeline/processors/digitizer.py
@@ -39,6 +39,10 @@ class DigitizerProcessor(_BaseProcessor):
         'digitizers.channel_1_B.apd.pulseIntegral': 'B',
         'digitizers.channel_1_C.apd.pulseIntegral': 'C',
         'digitizers.channel_1_D.apd.pulseIntegral': 'D',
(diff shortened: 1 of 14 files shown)


extra_foam/pipeline/processors/digitizer.py
bermudei, line 1: maybe just also keeping 'digitizers.channel_1_A.apd.pulseIntegral': 'A'. Just in case they need it

takluyver:
LGTM, thanks.

I haven't totally understood the hget related changes, but they look safe enough; I presume something changed with a new version of the redis client library.

bermudei (approved):
LGTM. The warning changes look also safe enough as thomas mentioned. Thanks James!

extra_foam/pipeline/processors/digitizer.py
JamesWrigley: Ah that's still in there :) Github just clips the file, if you press the  image button it should become visible.

JamesWrigley:
> I haven't totally understood the hget related changes, but they look safe enough; I presume something changed with a new version of the redis client library.

Yeah, so we're passing enums to the function and those used to be automagically converted to integers, but now it doesn't do that anymore so we have to call `.value` on the enum directly.

